PR #18, opened 2022-11-09
Setup Android app MVP with skeleton Welcome screen

crocsandcoffee:
## 📑 What does this PR do?

- Sets up the initial android app with a no-op Welcome Screen
- Minor fixes to different grade scripts to build the android app
- Fixes build issues with desktop related dependencies being used in common sources. For now I moved the affected files to a `desktopMain` source but we will want to fix move back shared composables and utility without the offending desktop dependencies

# ✅ Checklist

- [x] I have updated the documentation as required

- Checkout this branch, and run the gradle command in the README which will build and install the APK on your connected emulator or device

- Replace the placeholder UI with a polished welcome screen
- Hook up navigation graph for Login and Sign Up Screens
- Probably need to look at updating the Mastodon Theme so we can start using that throughout all composables.

Reminder - this screen is just a dummy UI. I just wanted to get everything up and running and commit the android setup to the repo


roshanrai06 (approved):

build.gradle.kts
```diff
@@ -1,18 +1,19 @@
-plugins {
-    `kotlin-dsl`
-}
-
```
crocsandcoffee, line 4: Removed - this is not needed.

build.gradle.kts
```diff
@@ -1,18 +1,19 @@
...
 buildscript {
+
+    extra["targetSDKVersion"] = 33
+    extra["compileSDKVersion"] = 33
+    extra["minSDKVersion"] = 23
```
crocsandcoffee, line 5: For now, I am placing versions via extras on the project. We can decide how/where we want to handle the versioning of dependencies across the entire project to ensure version parity

ui/welcome/src/commonMain/kotlin/social/androiddev/welcome/composables/WelcomeScreen.kt
```diff
@@ -0,0 +1,53 @@
...
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.unit.dp
+import social.androiddev.common.theme.MastodonTheme
+
+@Composable
+fun WelcomeScreen(
+    modifier: Modifier = Modifier,
+) {
+    MastodonTheme {
```
crocsandcoffee, line 20: 1) We should start using the `MastodonTheme` properly across all UI. This will open the door for someone else to tackle supporting Material You and building upon the theme as we focus on getting MVP functionality done...

2) Ignore this layout, it's just a dummy placeholder added to see if everything worked end-to-end. In a follow-up PR we can actually implement the full Welcome Screen with LogIn/SignUp flows.

build.gradle.kts
roshanrai06: do we want to keep min SDK to 23? it could be higher since API 24 or 26 covers more than 80%.

crocsandcoffee: Not sure.. I haven't thought that far. We can chat over it in slack and see what the group thinks. I don't feel strongly here. But, If we do want to increase adoption and availability, especially globally, then probably a minSDK of 6 is the way to go

build.gradle.kts
```diff
@@ -1,18 +1,19 @@
...
     }
 
     dependencies {
-        classpath("org.jetbrains.compose:compose-gradle-plugin:1.2.0")
-        classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:1.7.10")
+        classpath("org.jetbrains.compose:compose-gradle-plugin:1.2.1")
+        classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:1.7.20")
         classpath("com.android.tools.build:gradle:7.2.2")
-        classpath("org.jetbrains.kotlin:kotlin-serialization:1.7.10")
+        classpath("org.jetbrains.kotlin:kotlin-serialization:1.7.20")
```
thebino, line 16: I would not raise the versions in this PR. I've had some issues when updating, so stuck with these older versions.

thebino (changes requested):
Do not move the 
`ui/common/src/commonMain/kotlin/social/androiddev/common/composables/`
files into `desktopMain/`
They should be reusable from all platforms

ui/welcome/src/androidMain/AndroidManifest.xml
```diff
@@ -0,0 +1,3 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+          package="social.androiddev.welcome" />
```
thebino, line 3: should be `social.androiddev.ui.welcome`

ui/welcome/src/commonMain/kotlin/social/androiddev/welcome/composables/WelcomeScreen.kt
thebino: Let's move the `MastodonTheme` into the highest layer so the composable and screens are free to use in a separate app.

crocsandcoffee:
> Do not move the 
> `ui/common/src/commonMain/kotlin/social/androiddev/common/composables/`
> files into `desktopMain/`
> They should be reusable from all platforms

Let's chat offline. They are referencing desktop related dependencies and as such cause build issues with the android app.

I just moved them for the quick fix to unblock getting the android app working, but you'll need to rework those utils and composables to pull out the desktop related code outside of the common/shared source set.

ui/welcome/src/commonMain/kotlin/social/androiddev/welcome/composables/WelcomeScreen.kt
crocsandcoffee: Yup - take note of my comment. I just threw everything here to see it working end-to-end, including the theme. I'll be pulling that out and placing it at root in the next PR when I actually implement this screen.

build.gradle.kts
crocsandcoffee: This fixes warnings and (I think) an error I was getting initially with the android setup.

I'm guessing you probably copied over the templates for KMP which had outdated versions.. i tested running both desktop and the android app with this PR and these changes.

ui/welcome/src/androidMain/AndroidManifest.xml
racka98: These should all be moved to gradle file (namespace). They will no longer supported in future feature of AGP. AGP 7.3.x already throws a bunch of warnings about this

build.gradle.kts
racka98: minSDK 23 will cause a lot of headache with desugaring when KMP and kotlinx-datetime are involved.

thebino: Support as low as possible but try to reduce the maintenance only to cover legacy versions.
Always keep in mind, there are regions on earth, where not the latest phone is in everyone’s pocket.

I’m of course fine with raising to 24 or 26 as well. At least when we add video playback to the timeline, we’ll need to raise it most probably.

racka98: Also, I think using Versions catalog will be easier. 
I can make a PR adding them once this is merged

ui/welcome/src/androidMain/AndroidManifest.xml
crocsandcoffee: @racka98 mind doing that in your follow-up PR with the version catalog?

PR #43, opened 2022-11-10
Create new FeedCard composable

crocsandcoffee:
## 📑 What does this PR do?

- Added new FeedCard for MVP
- Replaced existing Timeline UI code. It was referencing domain model in the UI, and also for MVP we only will display immutable feed items in timeline, so no need to support actions and other account related info

# ✅ Checklist

Uncomment Preview or see screenshots

Screen Shot 2022-11-09 at 8 09 06 PM

Screen Shot 2022-11-09 at 8 09 15 PM

Screen Shot 2022-11-10 at 12 34 09 AM


ui/timeline/src/commonMain/kotlin/social/androiddev/timeline/FeedCard.kt
```diff
@@ -0,0 +1,217 @@
...
+                .padding(contentPadding),
+            verticalAlignment = Alignment.CenterVertically,
+        ) {
+            UserAvatar(
+                modifier = Modifier
+                    .size(72.dp)
+                    .clip(CircleShape)
+                    // TODO We should add support for a default image placeholder
+                    // in the case of loading or error
+                    .background(Color.Red),
```
thebino, line 89: this should be done inside the UserAvatar
Currently a placeholder is used. But I would change it to a composable param instead.

ui/timeline/src/commonMain/kotlin/social/androiddev/timeline/FeedCard.kt
```diff
@@ -0,0 +1,217 @@
...
+                userAddress = userAddress,
+                toot = toot,
+                date = date,
+            )
+        }
+    }
+}
+
+@Composable
+private fun TootTextContent(
```
thebino, line 107: This should definitely be extracted since there are multiple types of content (text only, image, video..)

crocsandcoffee: Yup! Was thinking about probably exposing something like `content: @Composable -> Unit` and based on the actual type (text only, image, video, multiple images), we would pass the corresponding component for each of those. i can do that additional work in this PR, no worries. Just wanted to keep it simple and small for now

ui/timeline/src/commonMain/kotlin/social/androiddev/timeline/FeedCard.kt
crocsandcoffee: Good to know -  I missed that we had that. Will Update next commit

ui/timeline/src/commonMain/kotlin/social/androiddev/timeline/FeedCard.kt
```diff
@@ -0,0 +1,217 @@
...
+import social.androiddev.common.theme.MastodonTheme
+
+/**
+ * Stateful
+ */
+@Composable
+fun FeedCard(
+    state: FeedItemState,
+    contentPadding: PaddingValues = PaddingValues(8.dp),
+    modifier: Modifier = Modifier,
```
racka98, line 39: I believe the `modifier` should start before `contentPadding`.  

According to [Modifier docs](https://developer.android.com/reference/kotlin/androidx/compose/ui/Modifier):

> Composables that accept a [Modifier](https://developer.android.com/reference/kotlin/androidx/compose/ui/Modifier) as a parameter to be applied to the whole component represented by the composable function should name the parameter modifier and assign the parameter a default value of [Modifier](https://developer.android.com/reference/kotlin/androidx/compose/ui/Modifier). It should appear as the first optional parameter in the parameter list; after all required parameters (except for trailing lambda parameters) but before any other parameters with default values. Any default modifiers desired by a composable function should come after the modifier parameter's value in the composable function's implementation, keeping [Modifier](https://developer.android.com/reference/kotlin/androidx/compose/ui/Modifier) as the default parameter value

crocsandcoffee: Yup - good catch. I thought I did this in all of them and must have missed it here. Maybe we can set up some rules to catch this as well.

racka98: The Detekt setup I am adding has all these rules. You can also check out [these rules](https://twitter.github.io/compose-rules/rules) compiled by Twitter

ui/timeline/src/commonMain/kotlin/social/androiddev/timeline/FeedCard.kt
```diff
@@ -0,0 +1,228 @@
...
+                date = date,
+                images = images,
+                videoUrl = videoUrl
+            )
+        }
+    }
+}
+
+@Composable
+private fun TootContent(
```
thebino, line 109: I would extract this in a separate file to reduce the size here

ui/timeline/src/commonMain/kotlin/social/androiddev/timeline/FeedCard.kt
```diff
@@ -0,0 +1,228 @@
...
+    Spacer(Modifier.width(12.dp))
+}
+
+@Composable
+fun VerticalSpacer() {
+    Spacer(Modifier.height(8.dp))
+}
+
+@Composable
+private fun TopLabel(
```
thebino, line 151: Also a separate file

ui/timeline/src/commonMain/kotlin/social/androiddev/timeline/TimelineScreen.kt
```diff
@@ -1,99 +1,46 @@
...
-                    Text(
-                        text = status.createdAt,
-                        color = Color(0xFF616981)
-                    )
-                }
             }
-            Text(status.content)
         }
     }
-    // TODO: footer (reply, boost, favorite, share)
```
thebino, line 68: Should we add the actions in here or below it in the parent item?

thebino (approved):
some minor adjustments, but nothing blocking

ui/timeline/src/commonMain/kotlin/social/androiddev/timeline/TimelineScreen.kt
crocsandcoffee: So I figured for MVP we won't support any actions, just Create an account/Log in -> View read-only Local Timeline and the ability to create a Toot (maybe just text initially?)

So I didn't want to add a bunch of TODOs and commented-out code here until we get to it.

ui/timeline/src/commonMain/kotlin/social/androiddev/timeline/FeedCard.kt
crocsandcoffee: Agreed - will move this and the TopLabel to a different file and rename to something more general.

PR #44, opened 2022-11-10
Create themed TextButton composable

crocsandcoffee:
## 📑 What does this PR do?

- Create wrapper Themed TextButton composable

# ✅ Checklist


ui/common/src/commonMain/kotlin/social/androiddev/common/composables/buttons/MastodonTextButton.kt
```diff
@@ -0,0 +1,156 @@
...
+    colors: ButtonColors = textButtonColors(),
+    onClick: () -> Unit,
+) {
+    TextButton(
+        modifier = modifier,
+        enabled = enabled,
+        colors = colors,
+        onClick = onClick
+    ) {
+        Text(text.uppercase())
```
thebino, line 38: Is this text always uppercase? maybe just put the text in here and the caller can decided what the text looks like?

thebino (approved):

ui/common/src/commonMain/kotlin/social/androiddev/common/composables/buttons/MastodonTextButton.kt
crocsandcoffee: Yes, all the button composables (MastodonButton, MastodonOutlinedButton) and this force upper case so it's consistent across all buttons in the app. I wanted to initially make this as part of the button style but that's not supported. 

I don't feel strongly, I can remove, but jut wanted to have some sort of consistency to the button text styling so we don't have Sentence Case and then ALL CAPS in different places.

PR #41, opened 2022-11-09
Add GitHub workflows for KMP

bitPogo:
## 📑 What does this PR do?

This addresses partially #12 by adding a setup which works for KMP.
It also adjusts the setups of KMP projects in order to make them properly work in the CI.

**This PR depends on #58**

# ✅ Checklist

- [x] My pull request adheres to the code style of this project
- [ ] My code requires changes to the documentation
- [ ] I have updated the documentation as required
- [ ] All the tests have passed

## 🧪 How can this PR been tested?

If the CI starts and runs consider it tested.

PR #58, opened 2022-11-11
Change UrlStream

bitPogo:
## 📑 What does this PR do?

This simply changes `UrlStream` to accommodate the fact that for native is (currently) no concept of an ByteArrayInputStream, which is completely blocking testing for iOS at the moment.

**This PR depends on #57** 

# ✅ Checklist

### DIFF
--- a/data/network/src/androidMain/kotlin/social/androiddev/common/network/UrlStream.kt
+++ b/data/network/src/androidMain/kotlin/social/androiddev/common/network/UrlStream.kt
@@ -6,6 +6,6 @@ import io.ktor.client.engine.cio.CIO
 import io.ktor.client.request.get
 import java.io.ByteArrayInputStream
 
-actual suspend fun urlStream(url: String): ByteArrayInputStream = HttpClient(CIO).use {
-    ByteArrayInputStream(it.get(url).body())
+suspend fun String.toUrlStream(): ByteArrayInputStream = HttpClient(CIO).use {
+    ByteArrayInputStream(it.get(this).body())
 }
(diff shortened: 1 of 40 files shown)


data/network/src/androidMain/kotlin/social/androiddev/common/network/UrlStream.kt
```diff
@@ -6,6 +6,6 @@ import io.ktor.client.engine.cio.CIO
 import io.ktor.client.request.get
 import java.io.ByteArrayInputStream
 
-actual suspend fun urlStream(url: String): ByteArrayInputStream = HttpClient(CIO).use {
-    ByteArrayInputStream(it.get(url).body())
+suspend fun String.toUrlStream(): ByteArrayInputStream = HttpClient(CIO).use {
+    ByteArrayInputStream(it.get(this).body())
```
crocsandcoffee, line 10: I think alternatively a better approach here would be to either:
A) Update the `urlStream` contract so it's available/implemented by android, iOS, and Desktop JVM
B) Remove the contract and create some platform-agnostic solution in `commonMain`

crocsandcoffee (changes requested):
Seems like this PR and #41 have the same/similar changes, and I have same concern below on both. Let's chat over slack to quickly hash this out to unblock you so we can get both PRs moving 👍🏼

data/network/src/androidMain/kotlin/social/androiddev/common/network/UrlStream.kt
thebino: This s what I'm trying to say. See #62 how I have done it.

PR #61, opened 2022-11-13
remove unnecessary blocking call in tests and introduce coroutine

thebino:
## 📑 What does this PR do?

remove unnecessary blocking call in tests and introduce coroutine tests. This was already added in #41 and #58 but should be a separate PR to reduce the changes and complexity of these PRs

# ✅ Checklist

### DIFF
--- a/data/network/src/commonTest/kotlin/social/androiddev/common/network/MastodonApiTests.kt
+++ b/data/network/src/commonTest/kotlin/social/androiddev/common/network/MastodonApiTests.kt
@@ -18,11 +18,15 @@ import kotlin.test.assertTrue
 import kotlinx.coroutines.runBlocking
 import kotlinx.serialization.json.Json
 import kotlin.test.Test
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.runTest
 
+@OptIn(ExperimentalCoroutinesApi::class)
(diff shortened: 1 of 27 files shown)


data/network/src/commonTest/kotlin/social/androiddev/common/network/MastodonApiTests.kt
```diff
@@ -18,11 +18,15 @@ import kotlin.test.assertTrue
 import kotlinx.coroutines.runBlocking
 import kotlinx.serialization.json.Json
 import kotlin.test.Test
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.runTest
 
+@OptIn(ExperimentalCoroutinesApi::class)
```
crocsandcoffee, line 24: Suggest adding this to kotlin compiler args in certain modules so we don't have to specify it all over the place

crocsandcoffee (approved):

data/network/src/commonTest/kotlin/social/androiddev/common/network/MastodonApiTests.kt
thebino: I tried, but it looks like it is not available in KMP. I'll investigate later and do a PR if it is possible.

PR #66, opened 2022-11-16
Add Spacers

hrules6872:
## 📑 What does this PR do?

Add Spacers to start creating our own design language

# ✅ Checklist


exallium:
This took me more than a while to reconcile with myself on, because when I see a bunch of top-level functions like this the first thing I think is that we're going to end up polluting our class-space very quickly.

Are there opinions on utilizing some kind of namespacing here?

```
object Spacers {
  @Composable
  fun XLarge(modifier: Modifier = Modifier) {} // ...
}

// Usage
Spacers.XLarge()
```

This has the upside of making it easier to find things. If I want a specific spacer I can do `Spacers.` and get a nice list.

I can be convinced either way, just a random musing.

ui/timeline/src/commonMain/kotlin/social/androiddev/timeline/TimelineCard.kt
```diff
@@ -10,14 +10,7 @@
...
 // import androidx.compose.desktop.ui.tooling.preview.Preview
-import androidx.compose.foundation.layout.Box
-import androidx.compose.foundation.layout.PaddingValues
-import androidx.compose.foundation.layout.Row
-import androidx.compose.foundation.layout.fillMaxWidth
-import androidx.compose.foundation.layout.padding
-import androidx.compose.foundation.layout.size
-import androidx.compose.foundation.layout.wrapContentHeight
-import androidx.compose.foundation.layout.wrapContentSize
+import androidx.compose.foundation.layout.*
```
thebino, line 13: Can we avoid wildcard imports?
https://sergiomartins8.medium.com/avoid-using-auto-asterisk-imports-on-intellij-idea-278b9a6f148c

crocsandcoffee: +1

ui/common/src/commonMain/kotlin/social/androiddev/common/composables/Spacers.kt
```diff
@@ -0,0 +1,64 @@
...
+}
+
+@Composable
+fun ColumnScope.WeightSpacer(modifier: Modifier = Modifier) {
+    Spacer(modifier = modifier.weight(1f))
+}
+
+@Composable
+fun XSmallSpacer(modifier: Modifier = Modifier) {
+    Spacer(modifier = modifier.size(2.dp))
```
crocsandcoffee, line 33: Agree with @exallium comment, and grouping these under a namespace really helps with discoverability when trying look up our component spacers. 

I was planning on doing something similar for common paddings like

```kotlin
object Spacing {
       val XS = 4.dp
       val S = 6.dp
       val L = 12.dp
       ...
}
```

ui/common/src/commonMain/kotlin/social/androiddev/common/composables/Spacers.kt
```diff
@@ -0,0 +1,64 @@
...
+}
+
+@Composable
+fun SmallSpacer(modifier: Modifier = Modifier) {
+    Spacer(modifier = modifier.size(4.dp))
+}
+
+@Composable
+fun DefaultSpacer(modifier: Modifier = Modifier) {
```
crocsandcoffee, line 42: I have a feeling every time I see DefaultSpacer used I'll need to look it up to see what the actual spacing is. Most likely if we are going to have our own design system we will use common spacings. Maybe we could rename these Spacers to something more descriptive.

Just some ideas:

`ElementSpacing()` // spacing between elements vertically or horizontally like an Avatar and the title after it
`ContainerSpacing()` // standard layout/parent/screen spacing 
...

ui/common/src/commonMain/kotlin/social/androiddev/common/composables/Spacers.kt
crocsandcoffee: Potentially yes, also I just wrote up a quick example. Updated my original comment with a different one that I think is more fitting.

And regardless, the discoverability comment still applies in general.

ui/common/src/commonMain/kotlin/social/androiddev/common/composables/Spacers.kt
```diff
@@ -0,0 +1,64 @@
...
+    Spacer(modifier = modifier.weight(1f))
+}
+
+@Composable
+fun ColumnScope.WeightSpacer(modifier: Modifier = Modifier) {
+    Spacer(modifier = modifier.weight(1f))
+}
+
+@Composable
+fun XSmallSpacer(modifier: Modifier = Modifier) {
```
Rafsanjani, line 32: I personally don't see anything wrong with using hardcoded DP values in compose for spacing instead of wrapping them in another composable, as long as the values conform to the 4 dp material grid sizing.

hrules6872: Keep in mind that this is a cross-platform project :) 
Hard-coded dimensions are not an option, even for a simple Android app, if you want it to be compatible with all kinds of devices.

ui/common/src/commonMain/kotlin/social/androiddev/common/composables/Spacers.kt
```diff
@@ -0,0 +1,64 @@
...
+import androidx.compose.foundation.layout.RowScope
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.size
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.unit.dp
+
+@Composable
+fun RowScope.WeightSpacer(modifier: Modifier = Modifier) {
+    Spacer(modifier = modifier.weight(1f))
```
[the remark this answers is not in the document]
hrules6872: You can't use `weight()` outside of those scopes. 
For me it has is more readable and cleaner this:
```
Row() {
  Text()
  WeightSpacer()
  Text()
}

```
than that:
```
Row() {
  Text()
  Spacer(modifier = modifier.weight(1f))
  Text()
}

```

ui/common/src/commonMain/kotlin/social/androiddev/common/composables/Spacers.kt
petitJAM, line 32: I'd recommend adding a doc comment with the actual spacing size in it to all of these.

```kotlin
/**
 * Size: `2.dp`
 */
@Composable
fun XSmallSpacer(...) {
  // ...
}
```

krizzu: Agreed, that'd come useful 👍

ui/common/src/commonMain/kotlin/social/androiddev/common/composables/Spacers.kt
krizzu: I feel like this collection of composables could grow in size with time, so maybe we could have it scoped into `Spacing` object, so that we have one go-to namespace?

Something like

```kotlin
object Spacing {
  @Composable
  fun Default() {
    Spacer(modifier = Modifier.size(2.dp))
  }

  @Composable
  fun XSmall() {
    Spacer(modifier = Modifier.size(1.dp))
  }
}
```

and usage:

```kotlin
@Composable
fun MyUi() {
  Spacing.Default()

  Spacing.XSmall()
}
```

ui/common/src/commonMain/kotlin/social/androiddev/common/composables/Spacers.kt
petitJAM: > [#66 (comment)](https://github.com/AndroidDev-social/MastodonCompose/pull/66#discussion_r1026354947)

I'm not sure what you're trying to say with this link? Using pixel dimensions is not good of course, but DP are totally fine (which is what you have here anyway).  Admittedly, I haven't worked very much with Compose Desktop, but I have used `dp` dimension values when I did.

hrules6872: This PR is a first approximation so I hardcoded all the dimensions.

Dimensions (like paddings) must be defined in the Theme in order to support all kind of devices/platforms.
So, we shouldn't put this kind of comments because they can be a lie :)

crocsandcoffee (approved):
It's not entirely clear to me what these custom Spacers are really solving, I think it might make more sense to have these implemented as modifiers instead.

However, I don't feel super strongly and I don't see any harm in getting them merged in for now. With that being said - I think adding some paddings to the theme and having these Spacers (or Modifiers if you choose to go that route) update to pull the padding from the theme instead of hardcoded `dp` would be a good POC to hopefully showcase the benefit a bit more.

Approving anyway, as it can be done in a follow-up PR. 

Thanks and nice work 🚀

hrules6872:
@crocsandcoffee I started to use that kind of custom Spacers in one of my side projects, so I thought it was a good and easy PR to start contributing to this project. Then it has become not as good and easy as I thought at the beginning 😅

The good thing is that I have started using namespaces for this kind of functions (as you have pointed out to me in the comments). **I am very grateful to learn from you guys** 🫶

That said, I have no strong feelings about this PR and I guess it's best to close it and forget it.

PR #67, opened 2022-11-16
Add MastodonTextField component

crocsandcoffee:
## 📑 What does this PR do?
- Adds wrapper around TextField composable for custom theming. This will be used in login/sign up screens

# ✅ Checklist

### DIFF
--- a/ui/common/src/commonMain/kotlin/social/androiddev/common/composables/text/MastodonTextField.kt
+++ b/ui/common/src/commonMain/kotlin/social/androiddev/common/composables/text/MastodonTextField.kt
@@ -1,3 +1,12 @@
+/*
+ * This file is part of MastodonX.
+ *
+ * MastodonX is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
+ *
+ * MastodonX is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along with MastodonX. If not, see <https://www.gnu.org/licenses/>.
+ */
 package social.androiddev.common.composables.text
 
 //import androidx.compose.desktop.ui.tooling.preview.Preview
--- a/ui/common/src/commonMain/kotlin/social/androiddev/common/composables/text/MastodonTextField.kt
+++ b/ui/common/src/commonMain/kotlin/social/androiddev/common/composables/text/MastodonTextField.kt
@@ -291,7 +291,7 @@ private class DefaultTextFieldColors(
 
 private const val AnimationDuration = 150
 
-//@androidx.compose.desktop.ui.tooling.preview.Preview
+// @androidx.compose.desktop.ui.tooling.preview.Preview
 @Composable
 fun PreviewTextField() {
     MastodonTheme {
--- a/ui/common/src/commonMain/kotlin/social/androiddev/common/composables/text/MastodonTextField.kt
+++ b/ui/common/src/commonMain/kotlin/social/androiddev/common/composables/text/MastodonTextField.kt
@@ -9,7 +9,6 @@
  */
 package social.androiddev.common.composables.text
 
-//import androidx.compose.desktop.ui.tooling.preview.Preview
 import androidx.compose.animation.animateColorAsState
 import androidx.compose.animation.core.tween
 import androidx.compose.foundation.interaction.InteractionSource
@@ -42,12 +41,12 @@ import social.androiddev.common.theme.MastodonTheme
 @Composable
 fun MastodonTextField(
     value: String,
+    onValueChange: (String) -> Unit,
     modifier: Modifier = Modifier,
     enabled: Boolean = true,
     isError: Boolean = false,
     singleLine: Boolean = false,
     maxLines: Int = Int.MAX_VALUE,
-    onValueChange: (String) -> Unit,
     label: @Composable (() -> Unit)? = null,
     placeholder: @Composable (() -> Unit)? = null,
     leadingIcon: @Composable (() -> Unit)? = null,
@@ -292,7 +291,7 @@ private class DefaultTextFieldColors(
 
 private const val AnimationDuration = 150
 
-//@Preview
+@androidx.compose.desktop.ui.tooling.preview.Preview
 @Composable
 fun PreviewTextField() {
     MastodonTheme {
--- a/ui/common/src/commonMain/kotlin/social/androiddev/common/composables/text/MastodonTextField.kt
+++ b/ui/common/src/commonMain/kotlin/social/androiddev/common/composables/text/MastodonTextField.kt
@@ -1,0 +1,301 @@
+package social.androiddev.common.composables.text
+
+//import androidx.compose.desktop.ui.tooling.preview.Preview
+import androidx.compose.animation.animateColorAsState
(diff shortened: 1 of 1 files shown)


ui/common/src/commonMain/kotlin/social/androiddev/common/composables/text/MastodonTextField.kt
```diff
@@ -0,0 +1,301 @@
+package social.androiddev.common.composables.text
+
+//import androidx.compose.desktop.ui.tooling.preview.Preview
```
exallium, line 3: I don't think we should make a habit of commenting imports?

ui/common/src/commonMain/kotlin/social/androiddev/common/composables/text/MastodonTextField.kt
```diff
@@ -0,0 +1,301 @@
...
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.unit.dp
+import social.androiddev.common.theme.MastodonTheme
+
+@Composable
+fun MastodonTextField(
+    value: String,
+    modifier: Modifier = Modifier,
```
exallium, line 36: Have you looked at the twitter rules for composables? I quite like it. It may be good to consider standardizing on something.

https://twitter.github.io/compose-rules/rules/

ui/common/src/commonMain/kotlin/social/androiddev/common/composables/text/MastodonTextField.kt
```diff
@@ -0,0 +1,301 @@
...
+        if (errorLabelColor != other.errorLabelColor) return false
+        if (placeholderColor != other.placeholderColor) return false
+        if (disabledPlaceholderColor != other.disabledPlaceholderColor) return false
+
+        return true
+    }
+
+    override fun hashCode(): Int {
+        var result = textColor.hashCode()
+        result = 31 * result + disabledTextColor.hashCode()
```
exallium, line 260: This is due to not having access to something like Objects.hash?

ui/common/src/commonMain/kotlin/social/androiddev/common/composables/text/MastodonTextField.kt
```diff
@@ -0,0 +1,301 @@
...
+        colors = colors,
+    )
+}
+
+/**
+ * Creates a [TextFieldColors] that represents the default input text, background and content
+ * (including label, placeholder, leading and trailing icons) colors used in a [TextField].
+ */
+@Composable
+private fun textFieldColors(
```
exallium, line 72: This feels like an unfortunate amount of boilerplate.

ui/common/src/commonMain/kotlin/social/androiddev/common/composables/text/MastodonTextField.kt
```diff
@@ -0,0 +1,301 @@
...
+        result = 31 * result + errorLabelColor.hashCode()
+        result = 31 * result + placeholderColor.hashCode()
+        result = 31 * result + disabledPlaceholderColor.hashCode()
+        return result
+    }
+}
+
+private const val AnimationDuration = 150
+
+//@Preview
```
exallium, line 286: What's the matter with committing previews?

ui/common/src/commonMain/kotlin/social/androiddev/common/composables/text/MastodonTextField.kt
crocsandcoffee: Unfortunately there is no KMP `@Preview` annotation supported atm. Only desktop. So it must be commented out for all these composables to work with KMP. 

I've been uncommenting it locally to test my previews/composables then commenting out when committing,

https://github.com/JetBrains/compose-jb/issues/1603

ui/common/src/commonMain/kotlin/social/androiddev/common/composables/text/MastodonTextField.kt
crocsandcoffee: Yes, I believe we have a detekt PR that adds this in 👍🏼

ui/common/src/commonMain/kotlin/social/androiddev/common/composables/text/MastodonTextField.kt
exallium: Are we able to move the annotation package to the preview annotation itself? Like

```
@androidx.compose.desktop.ui.tooling.preview.Preview
fun MyPreview() { ... }
```

This would help localize the preview to *one* spot in your file and if other people follow this pattern we wouldn't have to worry about accidentally forgetting to comment out an import line?

crocsandcoffee: I don't think so. Because this is a shared src. So building the android/ios apps will try to build this as well and the `@androidx.compose.desktop.ui.tooling.preview.Preview` is not available, so build will fail

ui/common/src/commonMain/kotlin/social/androiddev/common/composables/text/MastodonTextField.kt
crocsandcoffee: Yeah, I might end up removing this stuff later. Typically I like to wrap any composables (third party, including googles) that heavily pull from MaterialTheme under the hood. I've been creating a composable library with commonly used components for now and writing default colors until we work out the theme later on.

ui/common/src/commonMain/kotlin/social/androiddev/common/composables/text/MastodonTextField.kt
crocsandcoffee: https://github.com/AndroidDev-social/MastodonCompose/pull/67#discussion_r1024656705

ui/common/src/commonMain/kotlin/social/androiddev/common/composables/text/MastodonTextField.kt
crocsandcoffee: Not sure - I mimicked what the built-in `TextField` composable does under the hood. My guess is
`androidx.compose.ui.graphics.Compose` is not a data class so in order to fulfill the `@Immutable` contract for `DefaultTextFieldColors` we need manually calculate the hashCode for all color properties

ui/common/src/commonMain/kotlin/social/androiddev/common/composables/text/MastodonTextField.kt
exallium: Right, still comment it out, just use a fully qualified ref instead of an import so that there's only one block of code to comment.

crocsandcoffee: Ah I see what you're saying - yes agreed, I'll make that change moving forward here and elsewhere

ui/common/src/commonMain/kotlin/social/androiddev/common/composables/text/MastodonTextField.kt
thebino: But the modifier parameter should be always first

crocsandcoffee: If you're providing a default modifier, it should be the first in the list of defaults. Any params that don't have defaults should still be defined first. 

I'll make sure I run through all the composables I've added so far once Detekt PR is merged and fix up ordering of params where it's needed

exallium (approved):

thebino (approved):

PR #68, opened 2022-11-16
Setup skeleton of authenticating client app

crocsandcoffee:
## 📑 What does this PR do?

- I added the first example of a network call with more end-to-end architecture between different modules. This example is the authentication call done to register the client app, before taking them to OAuth for login.

At a high level here is how the changes look:
* Updated the `domain` module to not depend on `data` module. IMO it should be independent and only depend on platform agnostic shared utilities such as kotlin-utils. (Open for discussion!)
* Added new module under `data` for repository implementations that wrap network and persistence. data now has `network` , `persistence` and `repositories`
* `domain` module now has interfaces for repositories, and a new use case has been added that works directly with the repository interface.
* `data.network` has been updated to expose the relevant mastodon API for authenticating an application
* `data.repositories` has a new class that implements the AuthenticationRepository interface, and calls the API in `data.network` to perform the authentication request, then takes the response client metadata and will store in cache via a  `data.persistence` source (TODO added in PR here)

I would appreciate any thoughts, questions, feedback on the PR about the dependency graph and architecture set here in addition to whatever else. Thanks :slightly_smiling_face:

# ✅ Checklist


data/network/src/commonMain/kotlin/social/androiddev/common/network/util/RunCatching.kt
```diff
@@ -0,0 +1,46 @@
+package social.androiddev.common.network.util
```
crocsandcoffee, line 1: TODO: this file should get moved to a new standalone kotlin-utils module that any other module can depend on.

domain/timeline/build.gradle.kts
```diff
@@ -46,24 +46,20 @@ kotlin {
         // shared
 
         val commonMain by getting {
-            dependencies {
-                implementation(projects.data.network)
-            }
+            dependencies {}
```
crocsandcoffee, line 49: domain modules shouldn't depend on anything other than standalone utilities.

data/network/src/commonMain/kotlin/social/androiddev/common/network/MastodonApi.kt
```diff
@@ -18,4 +19,24 @@ interface MastodonApi {
      * @see https://docs.joinmastodon.org/methods/instance/
      */
     suspend fun getInstance(domain: String? = null): Result<Instance>
+
+
```
exallium, line 23: Should probably have an auto-formatter running on commit or something (extra line?)

data/network/src/commonMain/kotlin/social/androiddev/common/network/MastodonApiImpl.kt
```diff
@@ -13,12 +13,15 @@ import io.ktor.client.HttpClient
...
 import io.ktor.client.plugins.ResponseException
 import io.ktor.client.request.get
+import io.ktor.client.request.post
+import io.ktor.client.request.forms.formData
 import kotlinx.serialization.SerializationException
+import social.androiddev.common.network.model.Application
 import social.androiddev.common.network.model.Instance
+import social.androiddev.common.network.util.runCatchingIgnoreCancelled
 
 class MastodonApiImpl(
```
exallium, line 23: For discussion: https://octoperf.com/blog/2016/10/27/impl-classes-are-evil/#:~:text=Impl%20is%20meaningless&text=If%20you%20can%20find%20a,breaking%20the%20Single%20Responsibility%20principle.

In general, a public class called Impl is a bit of a smell to me. Are we intending to have multiple instances of this? Or is it solely here to allow for easier testing? If it is the latter, I'd like to recommend using a static constructor on `MastodonApi` and make this class `internal`

data/network/src/commonMain/kotlin/social/androiddev/common/network/MastodonApiImpl.kt
```diff
@@ -35,4 +38,25 @@ class MastodonApiImpl(
             Result.failure(exception = exception)
         }
     }
+
+    override suspend fun createApplication(
+        clientName: String,
+        redirectUris: String,
+        scopes: String,
+        website: String?
+    ): Result<Application> {
```
exallium, line 47: Is this the Kotlin result type? is it out of experimental internal apis now?

data/network/src/commonMain/kotlin/social/androiddev/common/network/model/Application.kt
```diff
@@ -18,12 +18,14 @@ import kotlinx.serialization.Serializable
 @Serializable
 data class Application(
     // required attributes
+    @SerialName("id") val id: String,
```
exallium, line 21: Neat! Assume this is JSON?

data/network/src/commonMain/kotlin/social/androiddev/common/network/util/RunCatching.kt
exallium: Rec something like `core-util`? Not sure how important the language name is here.

data/network/src/commonMain/kotlin/social/androiddev/common/network/util/RunCatching.kt
```diff
@@ -0,0 +1,46 @@
...
+
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.TimeoutCancellationException
+
+inline fun <reified T> runCatchingRetry(maxRetry: Int, block: () -> T): Result<T> {
+    var result: Result<T>
+    var retryCount = 0
+    do {
+        result = runCatching(block)
+        retryCount++
```
exallium, line 11: Something about these is a bit off-putting. Are these meant to run inside a coroutine?

Also, wondering if we want to enable a retry timeout via strategy pattern:

```
interface RetryTimeoutStrategy {
  fun getNextTimeout(lastTimeout: Long): Long
}

// Don't use this directly, you'll want to add a lower and upper bounds.
class ExponentialBackoffTimeoutStrategy {
  override fun getNextTimeout(lastTimeout: Long): Long = pow(lastTimeout, 2)
}
```

data/repository/src/commonMain/kotlin/social/androiddev/common/AuthenticationRepositoryImpl.kt
```diff
@@ -0,0 +1,32 @@
+package social.androiddev.common
+
+import social.androiddev.common.network.MastodonApi
+import social.androiddev.domain.authentication.repository.AuthenticationRepository
+
+class AuthenticationRepositoryImpl(
```
exallium, line 6: Same comment again re `internal` and a static creation method. We should do what we can to minimize API surfaces.

data/network/src/commonMain/kotlin/social/androiddev/common/network/util/RunCatching.kt
crocsandcoffee: I like it

data/network/src/commonMain/kotlin/social/androiddev/common/network/MastodonApi.kt
crocsandcoffee: Whoops, I will fix. Was hoping Detekt PR gets merged soon

data/network/src/commonMain/kotlin/social/androiddev/common/network/MastodonApiImpl.kt
crocsandcoffee: Yes it is - I think for now we have this as the wrapping return type but this may possibly change. I actually thought you couldn't use it as a return type but I guess that changed?

data/network/src/commonMain/kotlin/social/androiddev/common/network/model/Application.kt
crocsandcoffee: Yes, this is from `kotlinx.serialization`. I wonder if we don't need the annotation and by default, it uses reflection on the field name... I hate coupling all the data models with these annotations.. especially if we don't need them.

data/network/src/commonMain/kotlin/social/androiddev/common/network/util/RunCatching.kt
crocsandcoffee: I just added a bunch of common wrappers and inline functions that I use in different projects with `kotlin.Result`. I can remove these if we don't see value in them.

This isn't restricted to just coroutines, just to be able to inline a simple retry mechanism if needed with error catching

data/network/src/commonMain/kotlin/social/androiddev/common/network/model/Application.kt
thebino: Where is this `id` coming from? It is not found in the documentation: https://docs.joinmastodon.org/entities/application/

The annotation can be omitted if it is reflecting the name.
For `client_id` (snake_case) for example I would add it explicit since it clashes with the kotlin code style (camelCase)

data/network/src/commonMain/kotlin/social/androiddev/common/network/util/RunCatching.kt
thebino: I'm not a big fan of these extra modules. In big projects you'll end up in a really big module every module depends on.
With high-level functions and kotlin extensions these function can be reduced to a minimum anyway.

data/network/src/commonTest/kotlin/social/androiddev/common/network/MastodonApiTests.kt
```diff
@@ -74,14 +75,70 @@ class MastodonApiTests {
         assertNotNull(actual = result.getOrNull()?.uri)
     }
 
+    // TODO: fix loading json from resources
+    @Test
+    @Ignore
+    fun `create application request should fail with invalid response`() = runTest {
+        // TODO: Add test json
+        val content: String = ""
```
thebino, line 83: since the response is 9 lines, you can add it inline instead
https://docs.joinmastodon.org/methods/apps/#create-an-application

data/network/src/commonMain/kotlin/social/androiddev/common/network/util/RunCatching.kt
crocsandcoffee: Ok, I suppose they could just be moved to `domain` since most likely a lot of the other modules will depend on that anyway. I don't feel strongly over this but I do think over time we will find helpful functions/extensions/utils that would/could be used across the whole project.

data/network/src/commonMain/kotlin/social/androiddev/common/network/MastodonApiImpl.kt
thebino: I agree 100%
Afaik very less people understand this principle, that's why I started with the mostly known `*Impl` naming to have it as explicit and easy to understand for everyone. But since we finally decided for a DI framework and have a good start of a documentation, we can change it and redirect people to our documentation.

data/network/src/commonMain/kotlin/social/androiddev/common/network/model/Application.kt
crocsandcoffee: @thebino https://docs.joinmastodon.org/methods/apps/#create-an-application

data/network/src/commonMain/kotlin/social/androiddev/common/network/MastodonApiImpl.kt
crocsandcoffee: +1 - I was typing up a response to say similar. And for now I just copied this same pattern for the repository until we got DI setup etc. I didn't want to add even more changes to this PR beyond what I did. A good follow-up should be the DI framework though now that we agreed on the library. I can do that and then sweep through and fix both this and the Repository impl

data/repository/src/commonMain/kotlin/social/androiddev/common/AuthenticationRepositoryImpl.kt
crocsandcoffee: https://github.com/AndroidDev-social/MastodonCompose/pull/68#discussion_r1024762246

data/network/src/commonMain/kotlin/social/androiddev/common/network/model/Application.kt
thebino: Maybe the entity documentation is not up-2-date: https://docs.joinmastodon.org/entities/application/

data/network/src/commonTest/kotlin/social/androiddev/common/network/MastodonApiTests.kt
crocsandcoffee: Cool sounds good to me. Wasn't sure yet where we wanted to go with these tests for API calls but I can just in-line the actual JSON like you suggested here and update these two tests I added.

data/network/src/commonMain/kotlin/social/androiddev/common/network/util/RunCatching.kt
thebino: Yes, we might also have the domain layer splitted soon.
domain:authentication
domain:timeline
...

data/network/src/commonMain/kotlin/social/androiddev/common/network/MastodonApi.kt
```diff
@@ -9,6 +9,7 @@
  */
 package social.androiddev.common.network
 
+import social.androiddev.common.network.model.Application
```
crocsandcoffee, line 12: @thebino I just noticed the GPL license header is missing from files in `network` module. Is this intentional or should I add it? If so - I'll do it in follow-up PR

domain/timeline/build.gradle.kts
thebino: 👍

data/network/src/commonMain/kotlin/social/androiddev/common/network/model/Application.kt
```diff
@@ -18,12 +18,14 @@ import kotlinx.serialization.Serializable
...
     @SerialName("name") val name: String,
+
+    // client attributes
```
thebino, line 24: Not sure if these client attributes are also optional. Depends in which calls this entity is reused.

crocsandcoffee: these client attributes are exactly what we need from this response. We will need to cache/store these and use them in follow-up calls

data/network/src/commonMain/kotlin/social/androiddev/common/network/model/Application.kt
crocsandcoffee: yeah not sure - but I will confirm if `id` is returned from real call and update this model accordingly

data/network/src/commonMain/kotlin/social/androiddev/common/network/model/Application.kt
thebino: Yes, I know. But the `Application` entity is re-used for other responses without these client attributes.
Like in a status: https://docs.joinmastodon.org/methods/statuses/

So I would change them to be nullable and move it back to the bottom

crocsandcoffee: @thebino with regards to _data integrity_ - what are your thoughts on having separate models then in this case? It would make for cleaner/self-documenting models and structure, so I don't need to validate non-null outside of the deserialization

data/network/src/commonMain/kotlin/social/androiddev/common/network/MastodonApiImpl.kt
exallium: It was internal for a while but I guess they made it public. 

The unfortunate thing about Result is that your failure has to extend Throwable. So we can either keep result or write an `Either<L, R>` type which is quite simple thanks to `Nothing` and extension functions.

data/network/src/commonMain/kotlin/social/androiddev/common/network/util/RunCatching.kt
exallium: Yeah it just kinda looks like something someone could accidentally call from the main thread

data/network/src/commonMain/kotlin/social/androiddev/common/network/model/Application.kt
lotdrops: The `SerialName` annotation is only needed when the variable name is different to the attribute we parse. If we receive "id" I would remove the `SerialName`, but if we receive `client_id` I'd keep it to avoid having snake_case variable names

data/network/src/commonMain/kotlin/social/androiddev/common/network/MastodonApiImpl.kt
lotdrops: Personally I find _Either_ too abstract and I prefer _Result_, as it is clear about what it is and what it contains in each case/side. However, as I don't like using Throwables for the error side (I generally use enums and sealed classes), I have a custom result class (it can be further extended).  
What do you think? https://gist.github.com/lotdrops/cc0714ffc145f46d97539bf4b03fa050

crocsandcoffee: I was going to suggest something similar - as I don't think we should leak lower level errors/exceptions. These can be mapped to a type that the domain/app can consume and determine what to do if more granular handling is needed. 

Here is a gist of `APIResponse` I created, a return type for all API calls that maps the response code, error and body to a specific type.

https://gist.github.com/crocsandcoffee/4d88fde78bd61aeabbca4ab900deb210

data/network/src/commonMain/kotlin/social/androiddev/common/network/model/Application.kt
crocsandcoffee: Yes - I will do an isolated PR where I clean up the models and only add annotations to fields that are snake_case.

data/network/src/commonMain/kotlin/social/androiddev/common/network/util/RunCatching.kt
crocsandcoffee: Good observation @exallium  - I_can make this a `suspend` function to enforce that, and if the consumer really wanted to run on main thread they could use a wrapping Main dispatcher.

data/network/src/commonMain/kotlin/social/androiddev/common/network/MastodonApiImpl.kt
lotdrops: In my projects I use MyResult in all layers, and with the map utils I map errors and success from data to domain models, and from domain to UI models (when I have different UI models).   
For API, I create a MyResult.Error with `enum class GenericApiError { NoInternet, Network, Server, Unknown }`.
I also have another for specific errors:```

sealed class CompleteApiError<in T>
object NoInternetApiError : CompleteApiError<Any?>()
object NetworkApiError : CompleteApiError<Any?>()
class ServerApiError(
    override val message: String = "",
) : CompleteApiError<Any?>(), MessageError
object UnknownApiError : CompleteApiError<Any?>()
class ExpectedApiError<T>(
    val which: T,
    override val message: String = "",
) : CompleteApiError<T>(), MessageError
```
But if we want more specific classes yours looks good! We should change it to _ApiResponse_ though, shouldn't we?

**Comment by lotdrops** [comment]
Sorry if this was discussed elsewhere, I didn't find it. Are we settled on **modularization** by layer?  
UI needs to be separate from the rest because of KMP, that's for sure. However, I think we could split it by features instead of doing domain/data.  
As I understand architecture, we want modules with high cohesion and with low coupling. If we split by layer I think we do not achieve those goals. Also, modularization by feature mimics how we'd split by "teams". 
Also, I think it used to be common to modularize by layer but now I believe it is mostly done by feature. For example, it is shown in the [android architecture page](https://developer.android.com/topic/modularization/patterns). Fernando Cejas made it popular, but then [changed his mind](https://fernandocejas.com/2018/05/07/architecting-android-reloaded/#what-is-modularization) about this because of feedback from the community.
I usually follow an approach similar to [this](https://proandroiddev.com/modularization-of-android-applications-in-2021-a79a590d5e5b). For small/medium apps I don't split features in api/impl, and my module-injector is inside of app.

**Comment by crocsandcoffee** [comment]
> Sorry if this was discussed elsewhere, I didn't find it. Are we settled on **modularization** by layer? UI needs to be separate from the rest because of KMP, that's for sure. However, I think we could split it by features instead of doing domain/data. As I understand architecture, we want modules with high cohesion and with low coupling. If we split by layer I think we do not achieve those goals. Also, modularization by feature mimics how we'd split by "teams". Also, I think it used to be common to modularize by layer but now I believe it is mostly done by feature. For example, it is shown in the [android architecture page](https://developer.android.com/topic/modularization/patterns). Fernando Cejas made it popular, but then [changed his mind](https://fernandocejas.com/2018/05/07/architecting-android-reloaded/#what-is-modularization) about this because of feedback from the community. I usually follow an approach similar to [this](https://proandroiddev.com/modularization-of-android-applications-in-2021-a79a590d5e5b). For small/medium apps I don't split features in api/impl, and my module-injector is inside of app.

I think what we have in place right now is a mix of both, sort of.

We are doing somewhat feature modules. For example, right now we have `ui/timeline`, `ui/settings`, `ui/welcome`. Each of these feature modules contains the relevant UI/ViewModel/Screen(s)/Navigation. There is also a `ui/common` for shared UI components etc

We then have an ** optional ** domain/use-case layer that sits between these feature modules and data.

Our data module is also broken up into separate modules: `data/network`, `data/persistence`, and `data/repository`. Here is where it's more by "layer" than by feature.

So we are in a way doing a mix of modules broken down by features and by layer. I'm open to all and any suggestions, changes, etc. 

The purpose of this community-led project is to build something fun and useful, but also serve as a community-built example of good architecture, best practices, and bleeding-edge tech.

**Comment by lotdrops** [comment]
Yes, to this mix is what I refer to. The UI modularization looks good to me, it is with the data and domain layers where I'm not so sure about.
At least from a theoretical point of view, what makes more sense to me would be having modules such as feature/timeline, feature/authentication, feature/search, etc. And in those feature modules, I'd have both the domain and the data. Then, some core or common modules to provide with the setup and utils needed.
That said, I have used ktor and sqldelight very little, and I'm not sure how manageable it is to split them in modules. Maybe it is not doable and we have to stick with data/network and data/persistence...

**Comment by crocsandcoffee** [comment]
> Yes, to this mix is what I refer to. The UI modularization looks good to me, it is with the data and domain layers where I'm not so sure about. At least from a theoretical point of view, what makes more sense to me would be having modules such as feature/timeline, feature/authentication, feature/search, etc. And in those feature modules, I'd have both the domain and the data. Then, some core or common modules to provide with the setup and utils needed. That said, I have used ktor and sqldelight very little, and I'm not sure how manageable it is to split them in modules. Maybe it is not doable and we have to stick with data/network and data/persistence...

@lotdrops Yup I am in same boat here WRT experience with those two libraries and how all this will come together. Would you mind creating a discussion in the repo and link to your comment here so the community can give opinions/insight etc? For the time being I will continue with this as-is and we can also move things around once things become more clear or after getting more input. Thanks!

File: data/network/src/commonMain/kotlin/social/androiddev/common/network/MastodonApiImpl.kt (hunk 5 above)
Comment (crocsandcoffee) [reply data/network/src/commonMain/kotlin/social/androiddev/common/network/MastodonApiImpl.kt:47]: Yup sounds good! I will do a follow-up PR with this change in API layer to minimize the changes in this PR

File: data/network/src/commonMain/kotlin/social/androiddev/common/network/model/Application.kt (hunk 6 above)
Comment (crocsandcoffee) [reply data/network/src/commonMain/kotlin/social/androiddev/common/network/model/Application.kt:21]: OK, real JSON response shows `id`, so we should note that the documentation is not up-to-date moving forward.

Here's curL you can run if anyone is curious
```
curl -X POST \
	-F 'client_name=Test Application' \
	-F 'redirect_uris=urn:ietf:wg:oauth:2.0:oob' \
	-F 'scopes=read write follow push' \
	-F 'website=https://myapp.example' \
	https://mastodon.example/api/v1/apps
```

File: data/network/src/commonMain/kotlin/social/androiddev/common/network/util/RunCatching.kt (hunk 7 above)
Comment (crocsandcoffee) [reply data/network/src/commonMain/kotlin/social/androiddev/common/network/util/RunCatching.kt:11]: Added some javadoc comments for clarity, we may end up deleting/removing these if not needed. But if we do end up wanting to leverage some type of retry mechanism, we can look at bringing in something like your suggestion or tweaking these 👍🏼

File: data/network/src/commonMain/kotlin/social/androiddev/common/network/MastodonApiImpl.kt [hunk 12]
```diff
@@ -12,13 +12,15 @@ package social.androiddev.common.network
...
 import io.ktor.client.call.body
 import io.ktor.client.plugins.ResponseException
+import io.ktor.client.request.forms.formData
 import io.ktor.client.request.get
+import io.ktor.client.request.post
 import kotlinx.serialization.SerializationException
+import social.androiddev.common.network.model.Application
 import social.androiddev.common.network.model.Instance
-
+import social.androiddev.common.network.util.runCatchingIgnoreCancelled
```
exallium, line 21: some weird formatting here

data/network/src/commonMain/kotlin/social/androiddev/common/network/MastodonApiImpl.kt
exallium: Side note, <Any?> in these cases can be replaced using <Nothing> instead.

exallium (approved):
Looks good, assuming that the Impls will be internalized in a follow up (with the introduction of DI). Just a couple weird formatting comments I think.

PR #71, opened 2022-11-17
add git pre-commit hook

thebino:
## 📑 What does this PR do?

Adds a pre-commit script which can/should run before every commit to make it easier for devs to comply to the project rules and code styles

# ✅ Checklist

- [x] I have updated the documentation as required

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,0 +1,17 @@
+#!/bin/sh
+#
+# hook script to verify code formatting and runs tests for all modules
+# also it will try to format code and add missing license headers to every kotlin file.
+
+echo "***** Run code formatting and tests ******"
+git stash -q --keep-index
+
+./gradlew check
+
+status=$?
+
+git stash pop -q
+
+echo "***** Done with code formatting and tests ******"
+
+exit $status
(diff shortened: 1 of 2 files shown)


.githooks/pre-commit
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+#
```
crocsandcoffee, line 2: shouldn't all git hook scripts go under .git/hooks ?

.githooks/pre-commit
```diff
@@ -0,0 +1,17 @@
...
+# hook script to verify code formatting and runs tests for all modules
+# also it will try to format code and add missing license headers to every kotlin file.
+
+echo "***** Run code formatting and tests ******"
+git stash -q --keep-index
+
+./gradlew check
```
crocsandcoffee, line 9: shouldn't we also run `spotlessApply` or `spotlessCheck`?

What does `check` task run?

.githooks/pre-commit
thebino: after installing the git hook, you'll find it there. But `.git` is not saved in the git history itself, so it needs to be outside if you want to share it

.githooks/pre-commit
thebino: I added `spotlessApply` to the script.
`test` is running `build`, `lint`, `test` and `spotlessCheck` already

crocsandcoffee (approved):

PR #78, opened 2022-11-21
Initial Koin setup across modules

crocsandcoffee:
## 📑 What does this PR do?
- Renamed `MastodonApiImpl` to `MastodonApiKtor` and marked `internal`. Added Koin bean to provide the interface `MastodonApi` which encapsulates implementation.
- Marked `AuthenticationRepositoryImpl` as `internal`. Added Koin bean to provide the interface `AuthenticationRepository` which encapsulates implementation.
- Created network and repository koin modules
- Created new `di` module for shared DI as well as entry point for resolving all DI modules cross platform and cross project

# ✅ Checklist


thebino:
What benefit will it gain to have a dedicated `di` module? The modules can be placed in the regarding modules

crocsandcoffee:
> What benefit will it gain to have a dedicated `di` module? The modules can be placed in the regarding modules

@thebino my thinking was that if every module defines its own DI module, there needs to be some entry point that depends on _all_ modules in order to setup Koin properly for each respective application. And rather than have all the app modules (app-android, app-iOS, app-desktop) depend on _all_ modules, a dedicated `di` module that glues everything together would be better. For ex, the android-app shouldn't need to know or depend on `data:repository`, `data:network` and `data:persistance` modules. However it needs to include them in the `KoinApplication` that is created, and the dedicated `di` module handles that.

What are your thoughts?

thebino:
> > What benefit will it gain to have a dedicated `di` module? The modules can be placed in the regarding modules
> 
> @thebino my thinking was that with every module (`data:*`, `domain:*`, `feature:*`) defining its own DI module, there needs to be some entry point that depends on _all_ modules in order to setup Koin properly for each respective application. And rather than have all the app modules (app-android, app-iOS, app-desktop) depend on _all_ modules, a dedicated `di` module that glues everything together would be better. For ex, the `android-app` shouldn't need to know or depend on `data:repository`, `data:network` and `data:persistance` modules. However it needs to include them in the `KoinApplication` that is created, and the dedicated `di` module handles that.
> 
> What are your thoughts?

I think it is fine to have the modules inside of the regarding apps (android / desktop)
inside lower layers you don't need any dependency since you're always depending on the interface instead.

crocsandcoffee:
> > > What benefit will it gain to have a dedicated `di` module? The modules can be placed in the regarding modules
> > 
> > @thebino my thinking was that with every module (`data:*`, `domain:*`, `feature:*`) defining its own DI module, there needs to be some entry point that depends on _all_ modules in order to setup Koin properly for each respective application. And rather than have all the app modules (app-android, app-iOS, app-desktop) depend on _all_ modules, a dedicated `di` module that glues everything together would be better. For ex, the `android-app` shouldn't need to know or depend on `data:repository`, `data:network` and `data:persistance` modules. However it needs to include them in the `KoinApplication` that is created, and the dedicated `di` module handles that.
> > 
> > What are your thoughts?
> 
> I think it is fine to have the modules inside of the regarding apps (android / desktop)
> inside lower layers you don't need any dependency since you're always depending on the interface instead.

Hm, yeah. I did add app modules that are empty for now in case someone brought this up... 

 Couple things:

1) This makes me question if interfaces and their implementations should be in separate modules....

2) I just realized there's a function called `loadKoinModules` that we can use in each feature module, to dynamically add modules to the Koin context as needed. I'll look into that next.

data/network/src/commonMain/kotlin/social/androiddev/common/network/di/NetworkKoinModule.kt
```diff
@@ -12,18 +12,27 @@ package social.androiddev.common.network.di
...
 import io.ktor.http.URLProtocol
+import org.koin.core.module.Module
+import org.koin.core.module.dsl.singleOf
+import org.koin.dsl.module
+import social.androiddev.common.network.MastodonApi
+import social.androiddev.common.network.MastodonApiKtor
 
-// TODO Update with Koin
-object NetworkDIModule {
+val networkModule: Module = module {
```
exallium, line 21: Where do people stand on commenting publicly accessible API

data/network/src/commonMain/kotlin/social/androiddev/common/network/MastodonApiKtor.kt
```diff
@@ -15,11 +15,14 @@ import io.ktor.client.plugins.ResponseException
...
 import io.ktor.client.request.post
+import io.ktor.http.URLProtocol
+import io.ktor.http.path
 import kotlinx.serialization.SerializationException
 import social.androiddev.common.network.model.Application
 import social.androiddev.common.network.model.Instance
 import social.androiddev.common.network.util.runCatchingIgnoreCancelled
-class MastodonApiImpl(
+
+internal class MastodonApiKtor(
```
exallium, line 25: FTR I think `Impl` was likley fine as a name, as long as the class isn't leaking out of the module. You could also call it `RealMastodonApi`. Depends whether we want dependencies to be called out in class names. In this case maybe Ktor is ok, since we're injecting it with a class that comes from Ktor.

exallium (commented):
Have we discussed public API comments (those being any publicly accessible unit revealed between modules)

data/network/src/commonMain/kotlin/social/androiddev/common/network/di/NetworkKoinModule.kt
crocsandcoffee: I personally agree with that. Something that I missed as I am racing to get some structure in place. Will update with javadocs on all public APIs

data/network/src/commonMain/kotlin/social/androiddev/common/network/MastodonApiKtor.kt
crocsandcoffee: Yeah I thought to myself the suffix Ktor gives more value in naming than `Impl`, in that this client implementation is heavily coupled with Ktor. And if we wanted to swap out the client implementation with something else, it would become more clear than just recreating another "Impl" at that time and struggling with naming.

thebino (approved):

PR #82, opened 2022-11-22
Feature/adjust application requests

thebino:
## 📑 What does this PR do?

adjusted all application related requests with tests

# ✅ Checklist


data/network/src/commonMain/kotlin/social/androiddev/common/network/MastodonApiImpl.kt
```diff
@@ -58,4 +44,20 @@ class MastodonApiImpl(
                 }.body()
         }
     }
+
+    override suspend fun verifyApplication(): Result<Application> {
+        return runCatchingIgnoreCancelled<Application> {
+            httpClient.get("/api/v1/apps/verify_credentials") {
+                headers {
+                    append(HttpHeaders.Authorization, "testAuthorization")
```
crocsandcoffee, line 52: is this for debugging or the actual string that needs to be passed up?

data/network/src/commonMain/kotlin/social/androiddev/common/network/MastodonApiImpl.kt
```diff
@@ -58,4 +44,20 @@ class MastodonApiImpl(
...
+                }
+            }.body()
+        }
+    }
+
+    override suspend fun getInstance(domain: String?): Result<Instance> {
```
crocsandcoffee, line 58: `domain` shouldn't be null here I believe. but we can fix this later when we hook this up

data/network/src/commonMain/kotlin/social/androiddev/common/network/model/Application.kt
```diff
@@ -17,15 +17,24 @@ import kotlinx.serialization.Serializable
...
     val name: String,
+    @SerialName("vapid_key") val vapidKey: String,
+
+    // optional attributes
+    val website: String? = null,
+)
+
+@Serializable
+data class NewOauthApplication(
+    val id: String,
```
crocsandcoffee, line 30: probably good idea to break this out into it's own file with a javadoc on both of these models

crocsandcoffee (approved):

data/network/src/commonMain/kotlin/social/androiddev/common/network/model/Application.kt
thebino: javadoc yes, but a separate file will only bloat it.
In the API Doc the application is re-used. So you'll probably first have a look into this file

PR #99, opened 2022-12-13
fix(auth-storage): Thread safety

krizzu:
## 📑 What does this PR do?

Addresses https://github.com/AndroidDev-social/DodoForMastodon/issues/94

Making auth-storage thread safe with `mutex`. Also swapping Map with LinkedMap

# ✅ Checklist

### DIFF
--- a/data/persistence/src/commonMain/kotlin/social/androiddev/common/persistence/localstorage/DodoAuthStorage.kt
+++ b/data/persistence/src/commonMain/kotlin/social/androiddev/common/persistence/localstorage/DodoAuthStorage.kt
@@ -29,5 +29,5 @@ interface DodoAuthStorage {
     /**
      * Get the Access token for @param server
      */
-    suspend fun getAccessToken(server: String): String?
+    fun getAccessToken(server: String): String?
 }
--- a/data/persistence/src/commonMain/kotlin/social/androiddev/common/persistence/localstorage/DodoAuthStorage.kt
+++ b/data/persistence/src/commonMain/kotlin/social/androiddev/common/persistence/localstorage/DodoAuthStorage.kt
@@ -29,5 +29,5 @@ interface DodoAuthStorage {
     /**
      * Get the Access token for @param server
      */
-    fun getAccessToken(server: String): String?
+    suspend fun getAccessToken(server: String): String?
 }
--- a/data/persistence/src/commonMain/kotlin/social/androiddev/common/persistence/localstorage/DodoAuthStorageImpl.kt
+++ b/data/persistence/src/commonMain/kotlin/social/androiddev/common/persistence/localstorage/DodoAuthStorageImpl.kt
@@ -20,9 +20,9 @@ import kotlinx.serialization.json.Json
 
 internal class DodoAuthStorageImpl(
     private val settings: Settings,
-    private val json: Json
+    private val json: Json,
+    private val lock: Mutex = Mutex()
 ) : DodoAuthStorage {
-    private val lock = Mutex()
     override var currentDomain: String?
         get() = settings[KEY_DOMAIN_CACHE]
         set(value) {
--- a/data/persistence/src/commonMain/kotlin/social/androiddev/common/persistence/localstorage/DodoAuthStorageImpl.kt
+++ b/data/persistence/src/commonMain/kotlin/social/androiddev/common/persistence/localstorage/DodoAuthStorageImpl.kt
@@ -12,8 +12,9 @@ package social.androiddev.common.persistence.localstorage
 import com.russhwolf.settings.Settings
 import com.russhwolf.settings.get
 import com.russhwolf.settings.set
-import kotlinx.coroutines.sync.Mutex
-import kotlinx.coroutines.sync.withLock
+import kotlinx.atomicfu.locks.ReentrantLock
+import kotlinx.atomicfu.locks.reentrantLock
+import kotlinx.atomicfu.locks.withLock
 import kotlinx.serialization.Serializable
 import kotlinx.serialization.builtins.ListSerializer
 import kotlinx.serialization.json.Json
@@ -21,7 +22,7 @@ import kotlinx.serialization.json.Json
 internal class DodoAuthStorageImpl(
     private val settings: Settings,
     private val json: Json,
-    private val lock: Mutex = Mutex()
+    private val lock: ReentrantLock = reentrantLock()
 ) : DodoAuthStorage {
     override var currentDomain: String?
         get() = settings[KEY_DOMAIN_CACHE]
@@ -50,7 +51,7 @@ internal class DodoAuthStorageImpl(
         }
     private val memCache: LinkedHashMap<String, AccessToken> by lazy { diskCache }
 
-    override suspend fun getAccessToken(server: String): String? =
+    override fun getAccessToken(server: String): String? =
         lock.withLock { memCache[server]?.token }
 
     override suspend fun saveAccessToken(server: String, token: String) {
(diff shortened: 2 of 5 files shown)


data/persistence/src/commonMain/kotlin/social/androiddev/common/persistence/localstorage/DodoAuthStorageImpl.kt
```diff
@@ -20,7 +22,7 @@ internal class DodoAuthStorageImpl(
     private val settings: Settings,
     private val json: Json
 ) : DodoAuthStorage {
-
+    private val lock = Mutex()
```
crocsandcoffee, line 25: Should this be passed in through the constructor instead? Can always supply a default value in-line as well. We could then mock/fake easily with tests

data/persistence/src/commonMain/kotlin/social/androiddev/common/persistence/localstorage/DodoAuthStorageImpl.kt
```diff
@@ -31,29 +33,31 @@ internal class DodoAuthStorageImpl(
...
         get() {
             return settings
                 .getStringOrNull(KEY_ACCESS_TOKENS_CACHE)
                 ?.let { str ->
                     json.decodeFromString(ListSerializer(AccessToken.serializer()), str)
-                        .associateBy { it.server }
+                        .associateByTo(linkedMapOf()) { it.server }
                 }
-                ?: mutableMapOf()
+                ?: linkedMapOf()
```
crocsandcoffee, line 44: LinkedHashMap is mutable by default?

crocsandcoffee (approved):
Look's good, thanks for taking this off my shoulders 😄

data/persistence/src/commonMain/kotlin/social/androiddev/common/persistence/localstorage/DodoAuthStorageImpl.kt
krizzu: Yes - it extends from MutableMap (on JVM it's typealias to LinkedHashMap, which is also mutable)

data/persistence/src/commonMain/kotlin/social/androiddev/common/persistence/localstorage/DodoAuthStorageImpl.kt
krizzu: Yes, that's great suggestion!

krizzu:
Ready for merge

data/persistence/src/commonMain/kotlin/social/androiddev/common/persistence/localstorage/DodoAuthStorageImpl.kt
```diff
@@ -31,29 +33,31 @@ internal class DodoAuthStorageImpl(
...
             val list = value.map { it.value }
             settings[KEY_ACCESS_TOKENS_CACHE] =
                 json.encodeToString(ListSerializer(AccessToken.serializer()), list)
         }
+    private val memCache: LinkedHashMap<String, AccessToken> by lazy { diskCache }
 
-    private val memCache: MutableMap<String, AccessToken> by lazy { diskCache.toMutableMap() }
-
-    override fun getAccessToken(server: String): String? = memCache[server]?.token
+    override suspend fun getAccessToken(server: String): String? =
```
arkivanov, line 53: It appears that `suspend` is used only because of `Mutex`, which makes your calls async without a reason. You may try [atomicfu Locks](https://github.com/Kotlin/kotlinx-atomicfu#locks) instead, or write your own with expect/actual (it's pretty easy).

krizzu: Yeah, that does not seems right to have `suspend`able just because of that. Going to add atomicfu, I think it could come useful later on as well

data/repository/src/commonMain/kotlin/social/androiddev/common/repository/AuthenticationRepositoryImpl.kt
```diff
@@ -32,7 +32,7 @@ internal class AuthenticationRepositoryImpl(
         redirectUris: String,
         scopes: String,
         website: String?
-    ): NewAppOAuthToken? {
+    ): NewAppOAuthToken? = withContext(ioCoroutineContext) {
```
krizzu, line 35: I've noticed we're not making this `main` safe, so added withContext. There are no suspend points in those methods, however, so it'll fallback to sync anyway

data/persistence/src/commonMain/kotlin/social/androiddev/common/persistence/localstorage/DodoAuthStorage.kt
```diff
@@ -24,7 +24,7 @@ interface DodoAuthStorage {
     /**
      * Save the @param token keyed by @param server
      */
-    suspend fun saveAccessToken(server: String, token: String)
+    fun saveAccessToken(server: String, token: String)
```
krizzu, line 27: I also changed this method to be sync, as there are no suspend points

data/repository/src/commonMain/kotlin/social/androiddev/common/repository/AuthenticationRepositoryImpl.kt
arkivanov: My 50c here, `mastodonApi.createApplication` is already a `suspend` function, switching contexts should be performed on the callee site, not caller. E.g. Ktor as async by default.

krizzu: Aw yeah, I do agree with that - I'll create a follow up PR to inject context to ApiKtor client and switch context there

data/persistence/src/commonMain/kotlin/social/androiddev/common/persistence/localstorage/DodoAuthStorageImpl.kt
```diff
@@ -31,29 +34,31 @@ internal class DodoAuthStorageImpl(
...
-    private val memCache: MutableMap<String, AccessToken> by lazy { diskCache.toMutableMap() }
-
-    override fun getAccessToken(server: String): String? = memCache[server]?.token
+    override fun getAccessToken(server: String): String? =
+        lock.withLock { memCache[server]?.token }
 
-    override suspend fun saveAccessToken(server: String, token: String) {
-        memCache[server] = AccessToken(token = token, server = server)
-        diskCache = memCache
+    override fun saveAccessToken(server: String, token: String) {
```
crocsandcoffee, line 57: probably this should still be a `suspend` function. JSON encoding/decoding should probably not be done on main thread.

- [ ] Add back `suspend` for async saving of access token

data/persistence/src/commonMain/kotlin/social/androiddev/common/persistence/localstorage/DodoAuthStorage.kt
crocsandcoffee: I think this should be reverted back. Please see my other comment in review on more details

data/repository/src/commonMain/kotlin/social/androiddev/common/repository/AuthenticationRepositoryImpl.kt
crocsandcoffee, line 35: please revert these withContext calls in this file. Ktor is already async. We can make this additional change if we want in the Ktor client class but these changes shouldn't be needed

data/repository/src/commonMain/kotlin/social/androiddev/common/repository/AuthenticationRepositoryImpl.kt
```diff
@@ -102,10 +103,8 @@ internal class AuthenticationRepositoryImpl(
         }
     }
 
-    override suspend fun saveAccessToken(server: String, token: String) {
-        withContext(ioCoroutineContext) {
```
crocsandcoffee, line 106: please revert back

crocsandcoffee (changes requested):
Overall looks good! Please just make the saving of access token async again and remove the moot `withContext` changes

krizzu:
reverted back

krizzu:
changes reverted

crocsandcoffee (approved):

PR #102, opened 2022-12-15
feed data flow

digitalbuddha:
## 📑 What does this PR do?

# ✅ Checklist

Login and observe data rendering

Screen Shot 2022-12-18 at 1 45 59 PM


data/network/src/commonMain/kotlin/social/androiddev/common/network/MastodonApi.kt
```diff
@@ -77,4 +78,11 @@ interface MastodonApi {
      * @return an instance entity
      */
     suspend fun getInstance(domain: String? = null): Result<Instance>
+
+    /**
+     * Fetch home feed for a particular user
+     * @param accessToken representing the user
+     * @return a list of [Status]
+     */
```
crocsandcoffee, line 86: Can you add a link, similar to the other methods on this interface, to the official doc for this model

crocsandcoffee: Also, does the payload contain any other metadata such as "pageSize", "skip", "count" etc, for pagination support?

data/network/src/commonMain/kotlin/social/androiddev/common/network/MastodonApiKtor.kt
```diff
@@ -124,4 +125,21 @@ internal class MastodonApiKtor(
...
+                }
+            }.body<List<Status>>()
+            Result.success(
+                body
+            )
+        } catch (exception: SerializationException) {
+            Result.failure(exception = exception)
+        } catch (exception: ResponseException) {
+            Result.failure(exception = exception)
+        }    }
```
crocsandcoffee, line 144:
```suggestion

    override suspend fun getHomeFeed(domain: String, accessToken: String): Result<List<Status>> {
        return runCatchingIgnoreCancelled<<List<Status>> {
           httpClient.get("https://$domain/api/v1/timelines/home") {
                headers {
                    append(HttpHeaders.Authorization, "Bearer $accessToken")
                }
            }.body()
        }
```

data/network/src/commonMain/kotlin/social/androiddev/common/network/di/NetworkModule.kt
```diff
@@ -38,6 +38,14 @@ val networkModule: Module = module {
                     }
                 )
             }
+//            install(Auth) {
```
crocsandcoffee, line 41: 👍🏼

data/network/src/commonMain/kotlin/social/androiddev/common/network/model/Application.kt
```diff
@@ -17,9 +17,9 @@ import kotlinx.serialization.Serializable
  */
 @Serializable
 data class Application(
-    val id: String,
+    val id: String?=null,
```
crocsandcoffee, line 20: nullable `id` doesn't make sense to me... are you re-using this model for a different call which has nullable fields for some reason? 

If so, I would suggest creating a separate model for that case, so we don't lose this type safety. Otherwise, we will need to manually enforce the type safety on each call instead of a pure self-documenting model.

data/persistence/src/androidMain/kotlin/social/androiddev/common/persistence/di/AndroidPersistenceModule.kt
```diff
@@ -44,4 +45,13 @@ actual val persistenceModule: Module = module {
...
         AuthenticationDatabase(driver)
     }
+
+    single {
+        val driver = AndroidSqliteDriver(
+            schema = TimelineDatabase.Schema,
+            context = get(),
+            name = FEED_DB_NAME,
+        )
+        TimelineDatabase(driver)
```
crocsandcoffee, line 55: Do we want to cache the timeline? This is to support offline usage? Maybe it will become more clear later in the PR diff, but on the surface it seems this table/db will be REALLY large, especially for federated timeline. Also, is federated and local timeline stored in separate DB/tables?

data/persistence/src/commonMain/sqldelightTimeline/social/androiddev/common/timeline/Timeline.sq
```diff
@@ -0,0 +1,17 @@
+CREATE TABLE TimelineItem (
```
crocsandcoffee, line 1: Nice !

data/repository/build.gradle.kts
```diff
@@ -48,35 +48,43 @@ kotlin {
                 implementation(projects.domain.authentication)
                 implementation(libs.io.insert.koin.core)
                 implementation(libs.kotlinx.coroutines.core)
+                api(libs.store)
+                implementation ("com.squareup.sqldelight:coroutines-extensions:1.5.4")
```
crocsandcoffee, line 52: please update version catalog and reference library type-safe

data/repository/build.gradle.kts
```diff
@@ -48,35 +48,43 @@ kotlin {
...
+                implementation(libs.store)
+            }
         }
 
 
         // iOS
         val iosX64Main by getting
         val iosArm64Main by getting
-        val iosSimulatorArm64Main by getting
+//        val iosSimulatorArm64Main by getting
```
crocsandcoffee, line 77: is this temp? It's best to add some TODO or addition comment with context for any commented out code going into the repo so we don't need to track down PR, comments or discussions to understand why we commented out stuff

data/repository/src/commonMain/kotlin/social/androiddev/common/repository/di/RepositoryModule.kt
```diff
@@ -9,10 +9,23 @@
  */
 package social.androiddev.common.repository.di
 
+import com.squareup.sqldelight.runtime.coroutines.asFlow
+import com.squareup.sqldelight.runtime.coroutines.mapToList
 import kotlinx.coroutines.Dispatchers
 import org.koin.core.module.Module
 import org.koin.dsl.module
+import org.mobilenativefoundation.store.store5.Bookkeeper
```
crocsandcoffee, line 17: Was a bunch of stuff added to this module then removed?

data/repository/src/commonMain/kotlin/social/androiddev/common/repository/timeline/HomeTimelineRepository.kt
```diff
@@ -0,0 +1,37 @@
...
+import org.mobilenativefoundation.store.store5.Market
+import org.mobilenativefoundation.store.store5.MarketResponse
+import org.mobilenativefoundation.store.store5.ReadRequest
+import social.androiddev.common.timeline.TimelineItem
+
+interface HomeTimelineRepository {
+    suspend fun read(): Flow<MarketResponse<List<TimelineItem>>>
+}
+
+class RealHomeTimelineRepository(
```
crocsandcoffee, line 14: I think I like the "Real*" nomenclature over "*Impl". We should be consistent on this though throughout the project

data/repository/src/commonMain/kotlin/social/androiddev/common/repository/timeline/HomeTimelineRepository.kt
```diff
@@ -0,0 +1,37 @@
...
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.distinctUntilChanged
+import org.mobilenativefoundation.store.store5.Market
+import org.mobilenativefoundation.store.store5.MarketResponse
+import org.mobilenativefoundation.store.store5.ReadRequest
+import social.androiddev.common.timeline.TimelineItem
+
+interface HomeTimelineRepository {
+    suspend fun read(): Flow<MarketResponse<List<TimelineItem>>>
+}
```
crocsandcoffee, line 12: Unless we want to change this, the interface should be defined in `domain` layer/module. The ui/feature modules only depend on `domain` and not on any repository module directly. repository module(s) depend on domain module(s) and can then extend the relative interface

data/repository/src/commonMain/kotlin/social/androiddev/common/repository/timeline/TimelineRepoModule.kt
```diff
@@ -0,0 +1,125 @@
...
+import social.androiddev.common.timeline.TimelineDatabase
+import social.androiddev.common.timeline.TimelineItem
+
+/**
+ * Koin module containing all koin/bean definitions for
+ * timeline repository delegates.
+ */
+val timelineRepoModule: Module = module {
+
+    factory<HomeTimelineRepository> { RealHomeTimelineRepository(get()) }
```
crocsandcoffee, line 35: 👍🏼

gradle/libs.versions.toml
```diff
@@ -68,5 +70,5 @@ com-arkivanov-decompose-extensions-compose-jetbrains = { module = "com.arkivanov
 com-arkivanov-decompose-extensions-compose-jetpack = { module = "com.arkivanov.decompose:extensions-compose-jetpack", version.ref = "com-arkivanov-decompose" }
 multiplatform-settings = { module = "com.russhwolf:multiplatform-settings", version.ref = "com-russhwolf" }
 kotlinx-serialization-json = { module = "org.jetbrains.kotlinx:kotlinx-serialization-json", version.ref = "kotlinx-serialization" }
-
+store = {module = "org.mobilenativefoundation.store:store5", version.ref = "store"}
```
crocsandcoffee, line 73:
```suggestion
store = { module = "org.mobilenativefoundation.store:store5", version.ref = "store"}
```

ui/root/src/commonMain/kotlin/social/androiddev/root/navigation/DefaultRootComponent.kt
```diff
@@ -67,6 +68,7 @@ class DefaultRootComponent(
         componentContext: ComponentContext,
     ) = DefaultSignedInRootComponent(
         componentContext = componentContext,
+        mainContext=mainContext
```
crocsandcoffee, line 71:
```suggestion
        mainContext = mainContext
```

ui/signed-in/build.gradle.kts
```diff
@@ -33,6 +33,9 @@ android {
         }
     }
 }
+dependencies {
+    implementation(project(mapOf("path" to ":data:persistence")))
+}
```
crocsandcoffee, line 38: What's this for?

ui/signed-in/src/androidMain/AndroidManifest.xml
```diff
@@ -1,2 +1,4 @@
 <?xml version="1.0" encoding="utf-8"?>
-<manifest xmlns:android="http://schemas.android.com/apk/res/android"/>
\ No newline at end of file
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+    <uses-permission android:name="android.permission.INTERNET" />
```
crocsandcoffee, line 3: LOL the most famously forgotten about permission

ui/signed-in/src/commonMain/kotlin/social/androiddev/signedin/composables/SignedInRootContent.kt
```diff
@@ -56,10 +59,10 @@ fun SignedInRootContent(
 
 @Composable
 private fun TimelineTab(
-    component: TimelineComponent
+    state: StateFlow<MarketResponse<List<TimelineItem>>>
 ) {
     TimelineContent(
-        component = component,
+        state=state,
```
crocsandcoffee, line 65:
```suggestion
        state = state,
```

ui/timeline/build.gradle.kts
```diff
@@ -39,10 +39,13 @@ kotlin {
         val commonMain by getting {
             dependencies {
                 implementation(projects.domain.timeline)
+                implementation(projects.data.persistence)
+                implementation(projects.data.repository)
```
crocsandcoffee, line 43: I don't feel strongly, but we _should_ be consistent. The ui/* modules currently only should depend on domain/*, and not `data/*` directly. 

This does mean there needs to be a (sometimes redundant) UseCase/wrapper around the repository methods. But I like to think of it as: "If this view/screen/flow only needs access to a single method on my repository, should it really have access to all available APIs on the repo?"

Alternatively, if we move to a full end-to-end feature module architecture, then I am guessing the arch will look like:

app-android -> feature/timeline
feature/timeline -> domain/timeline, data/timeline, util/timeline, .... etc

ui/timeline/src/commonMain/kotlin/social/androiddev/timeline/TimelineContent.kt
```diff
@@ -29,10 +33,37 @@ import social.androiddev.timeline.navigation.TimelineComponent
...
 fun TimelineContent(
-    component: TimelineComponent,
+    state: StateFlow<MarketResponse<List<TimelineItem>>>,
     modifier: Modifier = Modifier,
 ) {
-    // TODO: Hook up to View Model for fetching timeline items
+   val items =  state.collectAsState()
+
+    when(val value = items.value){
+        is MarketResponse.Success->  {
```
crocsandcoffee, line 42:
```suggestion
        is MarketResponse.Success ->  {
```

ui/timeline/src/commonMain/kotlin/social/androiddev/timeline/TimelineContent.kt
```diff
@@ -29,10 +33,37 @@ import social.androiddev.timeline.navigation.TimelineComponent
...
+            val feedItems =  value.value.map {
```
crocsandcoffee, line 43:
```suggestion
            val feedItems = value.value.map {
```

ui/timeline/src/commonMain/kotlin/social/androiddev/timeline/TimelineContent.kt
```diff
@@ -29,10 +33,37 @@ import social.androiddev.timeline.navigation.TimelineComponent
...
+                dummyFeedItem.copy(
+                    id=it.statusId,
```
crocsandcoffee, line 45:
```suggestion
                    id = it.statusId,
```

ui/timeline/src/commonMain/kotlin/social/androiddev/timeline/navigation/TimelineComponent.kt
```diff
@@ -9,7 +9,67 @@
...
+
 /**
  * The base component describing all business logic needed for the timeline view
  */
-interface TimelineComponent
+interface TimelineComponent{
+    val state: StateFlow<MarketResponse<List<TimelineItem>>>
+}
+
+class DefaultTimelineComponent(
```
crocsandcoffee, line 38: I initially had these in same file, and got feedback to move the default impls to their own file (I forget the additional reasoning). But, let's be consistent with whatever we decide

ui/timeline/src/commonMain/kotlin/social/androiddev/timeline/navigation/TimelineComponent.kt
```diff
@@ -9,7 +9,67 @@
...
+    mainContext: CoroutineContext,
+    private val componentContext: ComponentContext,
+) : TimelineComponent, KoinComponent, ComponentContext by componentContext {
+
+    private val homeTimelineRepository:HomeTimelineRepository by inject()
```
crocsandcoffee, line 43: Nice yeah, this works for now and later I will sweep through to fix up all these components + VMs to being resolved via koin

ui/timeline/src/commonMain/kotlin/social/androiddev/timeline/navigation/TimelineComponent.kt
```diff
@@ -9,7 +9,67 @@
...
+            mainContext = mainContext,
+            homeTimelineRepository
+        )
+    }
+
+    override val state: StateFlow<MarketResponse<List<TimelineItem>>> = viewModel.state
+
+}
+
+class TimelineViewModel(
```
crocsandcoffee, line 56: maybe move to it's own file so this file doesn't become bloated?

ui/timeline/src/commonMain/kotlin/social/androiddev/timeline/navigation/TimelineComponent.kt
```diff
@@ -9,7 +9,67 @@
...
+    private val _state = MutableStateFlow<MarketResponse<List<TimelineItem>>>(MarketResponse.Empty)
+    val state: StateFlow<MarketResponse<List<TimelineItem>>> = _state.asStateFlow()
+    init {
+        scope.launch {
+            homeTimelineRepository.read().collect{
+               _state.value = it
+            }
+
+        }
+    }
```
crocsandcoffee, line 70:
```suggestion
    val state: StateFlow<MarketResponse<List<TimelineItem>>> = homeTimelineRepository
       .read()
       .mapLatest { it }
       .stateIn(scope, SharingStarted.Eagerly, MarketResponse.Empty)
```

data/repository/src/commonMain/kotlin/social/androiddev/common/repository/timeline/TimelineRepoModule.kt
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    factory {
+        NetworkFetcher.by(
+            get = { key: FeedType ->
+                when(key) {
+                    is FeedType.Home -> {
+                        val authStorage = get<DodoAuthStorage>()
+                        get<MastodonApi>()
+                            .getHomeFeed(authStorage.currentDomain!!, authStorage.getAccessToken(authStorage.currentDomain!!)!!)
```
crocsandcoffee, line 77: `!!` 😭  We'll fix this later

data/repository/src/commonMain/kotlin/social/androiddev/common/repository/timeline/TimelineRepoModule.kt
```diff
@@ -0,0 +1,125 @@
...
+ * Koin module containing all koin/bean definitions for
+ * timeline repository delegates.
+ */
+val timelineRepoModule: Module = module {
+
+    factory<HomeTimelineRepository> { RealHomeTimelineRepository(get()) }
+
+    factory<Store<FeedType, List<TimelineItem>, List<TimelineItem>>> {
+        val database = get<TimelineDatabase>()
+        Store.by(
```
crocsandcoffee, line 39: So Im not familiar with Store, but seems like there is a lot of logic going on in this koin module. Should a lot of the logic in this file be moved with a UseCase, which takes in a parameter(s) and then know's what feed to build/query?

data/repository/src/commonMain/kotlin/social/androiddev/common/repository/timeline/HomeTimelineRepository.kt
```diff
@@ -0,0 +1,37 @@
...
+    private val market:   Market<FeedType, List<TimelineItem>, List<TimelineItem>>
+) : HomeTimelineRepository {
+    /**
+     * returns a flow of home feed items from a database
+     * anytime table rows are created/updated will return a new list of timeline items
+     * on first return will also call network fetcher to get
+     * latest from network and update local storage with it]
+     */
+
+
```
crocsandcoffee, line 24:
```suggestion
```

data/repository/src/commonMain/kotlin/social/androiddev/common/repository/timeline/HomeTimelineRepository.kt
```diff
@@ -0,0 +1,37 @@
...
+    override suspend fun read(): Flow<MarketResponse<List<TimelineItem>>> {
+        return market.read(ReadRequest.of(
+            FeedType.Home,
+            emptyList(),
+            null,
+            true))
+            .distinctUntilChanged()
+    }
+
+
```
crocsandcoffee, line 34:
```suggestion
```

data/repository/src/commonMain/kotlin/social/androiddev/common/repository/timeline/HomeTimelineRepository.kt
```diff
@@ -0,0 +1,37 @@
...
+}
+
```
crocsandcoffee, line 36:
```suggestion
```

data/persistence/src/commonMain/sqldelightTimeline/social/androiddev/common/timeline/ConflictResolution.sq
```diff
@@ -0,0 +1,19 @@
+CREATE TABLE failedWrite (
```
crocsandcoffee, line 1: What's this file/table for?

crocsandcoffee (commented):
Look's super good! Nice first PR 👏🏼 👏🏼 👏🏼 

Don't forget to run `gradlew spotlessApply` and/or `gradlew check` locally, which should fix up a lot of the code style/formatting stuff and make sure tests didn't break!

data/network/build.gradle.kts
```diff
@@ -72,13 +73,11 @@ kotlin {
         // iOS
         val iosX64Main by getting
         val iosArm64Main by getting
-        val iosSimulatorArm64Main by getting
```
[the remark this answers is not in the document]
digitalbuddha: I was not targeting this in the store lib kmp targets. If we need it I can release a new version. I'm not sure what was using it but happy to add back.

data/network/src/commonMain/kotlin/social/androiddev/common/network/MastodonApi.kt
digitalbuddha: Absolutely, I'll add the rest of the fields we need as well. This was just getting end to end working, definitely in draft so far (forgot to mark as draft)

data/network/src/commonMain/kotlin/social/androiddev/common/network/di/NetworkModule.kt
digitalbuddha: yeah I wasn't sure what to do with this, I usually did auth through okhttp and need to learn how ktor wants to handle it. For now was just passing manually from repository

data/network/src/commonMain/kotlin/social/androiddev/common/network/model/Application.kt
digitalbuddha: Was rather weird to me too but the api returned null for an ID for some of the applications. We crashed on first fetch of my feed in androiddev.social. I'll make a localID for this for us and analyze why it is null. I'm not clear what Application even does for now

data/repository/src/commonMain/kotlin/social/androiddev/common/repository/timeline/TimelineRepoModule.kt
digitalbuddha: I'll move the logic to extension functions/static functions that we can test on their own. I'll inject single use cases/delegates here. For example the reader should just be something like `get<TimelineQueries>().feedFor(feedType)`

Pardon the dust and thank you for the callout. 

I'll add tests everywhere before landing this

data/repository/src/commonMain/kotlin/social/androiddev/common/repository/timeline/TimelineRepoModule.kt
digitalbuddha: ha yeah, really I'd love a flow<Pair<Domain, Token>() that I can switchmap from. This way if the user changes we automatically emit a new feed

data/persistence/src/commonMain/sqldelightTimeline/social/androiddev/common/timeline/ConflictResolution.sq
digitalbuddha: When we start posting toots, we want to save failures (no network) so that we can retry them. This way we can eagerly  resolve show new local data and save it to DB without having to post to API. When network is back and we make a read we want to resolve any conflicts we might have. This table keeps track of keys/toots that we may have a conflict with. 

For now just a placeholder until I add the updating/posting logic. BTW this is based on the new androidx offline first guidance

data/network/src/commonMain/kotlin/social/androiddev/common/network/MastodonApiKtor.kt
```diff
@@ -124,4 +125,13 @@ internal class MastodonApiKtor(
             Result.failure(exception = exception)
         }
     }
+
+    override suspend fun getHomeFeed(domain: String, accessToken: String): Result<List<Status>> {
+        return runCatchingIgnoreCancelled<<List<Status>> {
```
crocsandcoffee, line 130:
```suggestion
        return runCatchingIgnoreCancelled<List<Status>> {
```

crocsandcoffee: Sorry, this is my fault 🙈

data/persistence/src/androidMain/kotlin/social/androiddev/common/persistence/di/AndroidPersistenceModule.kt
digitalbuddha: Yup the goal is to be an offline first app and cache aggressively. We can expire/cleanup the DB table if it gets large but TBH feed data should be pretty small. 

As for storing together or separate. My current thinking is to have a column for FeedType when pulling posts/statuses and store in the same table but I haven't given data modeling much thought, open to any and all suggestions <3

data/repository/build.gradle.kts
digitalbuddha: will do before landing! thanks for catching

data/repository/build.gradle.kts
digitalbuddha: temp, will fix before landing

data/repository/src/commonMain/kotlin/social/androiddev/common/repository/di/RepositoryModule.kt
digitalbuddha: yes, I made my own module for when we move to feature gradle modules, let me cleanup

data/repository/src/commonMain/kotlin/social/androiddev/common/repository/timeline/HomeTimelineRepository.kt
digitalbuddha: Yeah I wanted to bring up, I followed square libraries that did this years ago and kinda stuck with it. Let's discuss in slack and if folks are fine I can refactor everything we have. 

Real is nice because you can have:
RealFoo
FakeFoo
UserFoo 
etc

data/repository/src/commonMain/kotlin/social/androiddev/common/repository/timeline/HomeTimelineRepository.kt
digitalbuddha: Sorry not following, If the real impl is in the data/repository layer how can the interface be in the domain layer?

ui/signed-in/build.gradle.kts
digitalbuddha: was studio auto adding dependency and not well. will remove

ui/timeline/build.gradle.kts
digitalbuddha: sure I can make the middle domain layer for consistency

data/repository/src/commonMain/kotlin/social/androiddev/common/repository/timeline/HomeTimelineRepository.kt
crocsandcoffee: yup, that's the pattern I am seeing right now and also growing very fond of it 😄

data/repository/src/commonMain/kotlin/social/androiddev/common/repository/timeline/HomeTimelineRepository.kt
crocsandcoffee: so `data/repo*` _depends on_ `domain`
`domain` contains: UseCases/simple repo wrappers, domain models, and interfaces for repositories
`data/repo*` provides the impls for the interfaces in `domain`
`data/repo*` has a koin module which binds the impl to the interface

ui/feature _depends on_ `domain` (not data/repo) (i.e. the ViewModels, Components, etc have the UseCases/Repo wrappers injected and do not have access to any data/repo)

data/network/build.gradle.kts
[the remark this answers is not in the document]
krizzu: This is for anyone using Apple Silicon, to run app locally (like iosX64 is simulator for intel). We don't support iOS for now, so I say it's fine to go

data/persistence/build.gradle.kts
```diff
@@ -12,6 +12,10 @@ sqldelight {
         packageName = "social.androiddev.common.persistence"
         sourceFolders = listOf("sqldelight")
     }
+    database("TimelineDatabase") {
```
krizzu, line 15: shouldn't this be collocated with a timeline feature, i.e `ui/timeline/src(...)/data` ?

data/network/build.gradle.kts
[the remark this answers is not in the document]
digitalbuddha: cool I can add back in, I need to get a diff out to update store. Will get done this week

data/network/src/commonMain/kotlin/social/androiddev/common/network/MastodonApi.kt
digitalbuddha: yeah there is pagination support through
```
max_id
String. Return results older than ID.
since_id
String. Return results newer than ID.
min_id
String. Return results immediately newer than ID.
limit
Integer. Maximum number of results to return. Defaults to 20 statuses. Max 40 statuses.
```

Added comment as well

data/network/src/commonMain/kotlin/social/androiddev/common/network/model/Application.kt
digitalbuddha: double checked docs and ID does not exist as a field https://docs.joinmastodon.org/entities/Application/
We can do a local ID in db

data/persistence/build.gradle.kts
digitalbuddha: I think there should be feature/data and feature/ui for most features, let me do as a follow up if everyone agrees

settings.gradle.kts
```diff
@@ -15,6 +15,7 @@ dependencyResolutionManagement {
         mavenCentral()
         mavenLocal()
         maven("https://maven.pkg.jetbrains.space/public/p/compose/dev")
+        maven("https://s01.oss.sonatype.org/content/repositories/snapshots")
```
digitalbuddha, line 18: Temp. I need to publish new version of store.

ui/signed-out/src/commonMain/kotlin/social/androiddev/signedout/selectserver/SelectServerContent.kt
```diff
@@ -52,7 +52,7 @@ fun SelectServerContent(
     onServerSelected: (String) -> Unit,
 ) {
 
-    var server by rememberSaveable { mutableStateOf("") }
+    var server by rememberSaveable { mutableStateOf("androiddev.social") }
```
digitalbuddha, line 55: Temp to ease debugging

ui/timeline/src/commonMain/kotlin/social/androiddev/timeline/TimelineContent.kt
```diff
@@ -29,14 +33,35 @@ import social.androiddev.timeline.navigation.TimelineComponent
...
-    TimelineContent(
-        items = listOf(dummyFeedItem),
-        modifier = modifier,
-    )
+   val items =  state.collectAsState()
+
+   val feedItems =  when(val value = items.value){
+        is StoreResponse.Data->  {
+            val feedItems =  value.value.map {
+               FeedItemState(
```
digitalbuddha, line 44: I'll move this to the VM

data/persistence/src/androidMain/kotlin/social/androiddev/common/persistence/di/AndroidPersistenceModule.kt
crocsandcoffee: I see ok, so we would just be caching some subset of the entire timeline, and having that expire/update as user scrolls when online? Might be worth considering separating the table for federated vs home timelines. Then you don't need to do a query based on some filter each time, as well as potentially get into a case where some fields are only present on federated vs home timeline and having to enforce type-safety when mapping DB -> domain models based on a FeedType etc

data/persistence/src/commonMain/sqldelightTimeline/social/androiddev/common/timeline/ConflictResolution.sq
crocsandcoffee: Got it, I dig it.

data/persistence/src/commonMain/sqldelightTimeline/social/androiddev/common/timeline/Timeline.sq
```diff
@@ -0,0 +1,30 @@
...
+  applicationName Text NOT NULL,
+  userName Text NOT NULL,
+  repliesCount INTEGER,
+  favouritesCount INTEGER,
+  reblogsCount INTEGER
+);
+
+insertFeedItem:
+INSERT OR REPLACE INTO TimelineItem
+VALUES (?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?);
```
crocsandcoffee, line 22: I think it's safer to use `?;` instead of `(?,?,?,?,?,?,?,?,?,?,?,?,?,?,?,?);` and insert via the TimelineItem model.

data/repository/build.gradle.kts
```diff
@@ -46,37 +46,45 @@ kotlin {
                 implementation(projects.data.network)
                 implementation(projects.data.persistence)
                 implementation(projects.domain.authentication)
+                implementation(projects.domain.timeline)
                 implementation(libs.io.insert.koin.core)
                 implementation(libs.kotlinx.coroutines.core)
+                //temp until we map to UI models
+                api(libs.store)
```
crocsandcoffee, line 53: Please add a TODO for easier visibility/discoverability later

data/repository/build.gradle.kts
```diff
@@ -46,37 +46,45 @@ kotlin {
...
             }
         }
 
 
         // android
         getByName("androidMain") {
             dependsOn(commonMain)
-            dependencies {}
+            dependencies {
+                api ("org.jetbrains.kotlinx:atomicfu:0.18.5")
```
crocsandcoffee, line 63: what's this for? can you add comment and/or add to version catalog?

data/repository/src/commonMain/kotlin/social/androiddev/common/repository/timeline/HomeTimelineRepository.kt
```diff
@@ -0,0 +1,32 @@
...
+class RealHomeTimelineRepository(
+    private val store: Store<FeedType, List<StatusUI>>
+) : HomeTimelineRepository {
+    /**
+     * returns a flow of home feed items from a database
+     * anytime table rows are created/updated will return a new list of timeline items
+     * on first return will also call network fetcher to get
+     * latest from network and update local storage with it]
+     */
```
crocsandcoffee, line 21: Nit -  this should be moved to the method on Interface.

domain/timeline/src/commonMain/kotlin/social/androiddev/domain/timeline/HomeTimelineRepository.kt
```diff
@@ -0,0 +1,13 @@
...
+import kotlinx.coroutines.flow.Flow
+
+import org.mobilenativefoundation.store.store5.StoreResponse
+import social.androiddev.domain.timeline.model.StatusUI
+
+interface HomeTimelineRepository {
+    suspend fun read(feedType: FeedType, refresh: Boolean = false): Flow<StoreResponse<List<StatusUI>>>
+}
+sealed class FeedType(val type:String){
+    object Home: FeedType("HOME")
```
crocsandcoffee, line 12: I think the domain model `FeedType` could just be an Enum. And you can map this Enum -> JSON string that needs to be passed to the API in the RealHomeTimelineRepostory. 

Just a thought, I don't feel strongly.

domain/timeline/src/commonMain/kotlin/social/androiddev/domain/timeline/model/StatusUI.kt
```diff
@@ -9,15 +9,21 @@
...
+data class StatusUI(
+    val remoteId:String,
+    val feedType: FeedType,
     val createdAt: String,
-    val repliesCount: Int,
-    val reblogsCount: Int,
-    val favouritesCount: Int,
+    val repliesCount: Long?,
+    val reblogsCount: Long?,
+    val favoritesCount: Long?,
```
crocsandcoffee, line 20: Shouldn't these just default to 0?

domain/timeline/src/commonMain/kotlin/social/androiddev/domain/timeline/model/StatusUI.kt
```diff
@@ -9,15 +9,21 @@
...
     val content: String,
-    val account: Account,
+    val account: Account?=null,
     val sensitive: Boolean = false,
     val spoilerText: String? = null,
-    val visibility: Visibility = Visibility.PUBLIC
+    val visibility: String = "Public",
```
crocsandcoffee, line 25: Maybe an Enum would be safer for visibility

domain/timeline/src/commonMain/kotlin/social/androiddev/domain/timeline/model/StatusUI.kt
```diff
@@ -9,15 +9,21 @@
  */
 package social.androiddev.domain.timeline.model
 
-data class Status(
-    val id: String,
+import social.androiddev.domain.timeline.FeedType
+
+data class StatusUI(
```
crocsandcoffee, line 14: Should "UI" models really exist in the domain layer? Or is this just a naming convention? 

I've been bitten in the past before when reference domain models in the UI layer itself. (And instead having a ViewModel or mapper transform domain model -> view/ui specific model.

ui/root/src/commonMain/kotlin/social/androiddev/root/navigation/DefaultRootComponent.kt
```diff
@@ -18,6 +18,7 @@ import com.arkivanov.decompose.router.stack.replaceCurrent
 import com.arkivanov.decompose.value.Value
 import com.arkivanov.essenty.parcelable.Parcelable
 import com.arkivanov.essenty.parcelable.Parcelize
+import social.androiddev.root.navigation.DefaultRootComponent.Config
```
crocsandcoffee, line 21: Is this an unused import?

ui/signed-out/src/commonMain/kotlin/social/androiddev/signedout/selectserver/SelectServerContent.kt
crocsandcoffee: Works for me.. I would add a TODO if you end up doing other commits, just for discoverability later if we want to do. quick search on what stuff is left to address

ui/timeline/src/commonMain/kotlin/social/androiddev/timeline/TimelineContent.kt
crocsandcoffee: Yup good call

ui/timeline/src/commonMain/kotlin/social/androiddev/timeline/TootContent.kt
```diff
@@ -45,14 +49,18 @@ fun TootContent(
         if (message != null) {
             Text(
                 modifier = Modifier.fillMaxWidth(),
-                text = message,
+                text = buildAnnotatedString {
+                    append(message)
```
crocsandcoffee, line 53: were you appending multiple text at some point? Shouldn' this just be `text= message`

ui/timeline/src/commonMain/kotlin/social/androiddev/timeline/navigation/TimelineComponent.kt
```diff
@@ -1,15 +1,12 @@
-/*
- * This file is part of Dodo.
- *
- * Dodo is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
- *
- * Dodo is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
- *
- * You should have received a copy of the GNU General Public License along with Dodo. If not, see <https://www.gnu.org/licenses/>.
- */
```
crocsandcoffee, line 9: You will want this added back. Can also just run `gradlew spotlessApply` so it updates all files for you

ui/timeline/src/commonMain/kotlin/social/androiddev/timeline/navigation/TimelineViewModel.kt
```diff
@@ -0,0 +1,41 @@
...
+    private val mainContext: CoroutineContext,
+    private val homeTimelineRepository: HomeTimelineRepository
+) : InstanceKeeper.Instance {
+    private val scope = CoroutineScope(mainContext + SupervisorJob())
+    private val _state = MutableStateFlow<StoreResponse<List<StatusUI>>>(StoreResponse.Loading(ResponseOrigin.SourceOfTruth))
+    val state: StateFlow<StoreResponse<List<StatusUI>>> = _state.asStateFlow()
+    init {
+        scope.launch {
+           try {
+               homeTimelineRepository.read(FeedType.Home, refresh = true).collect{
```
crocsandcoffee, line 28: FeedType should probably be passed into this ViewModel, so this can re-used for both home and federated timelines. There most likely shouldn't be a difference for them.

ui/timeline/src/commonMain/kotlin/social/androiddev/timeline/navigation/TimelineViewModel.kt
```diff
@@ -0,0 +1,41 @@
...
+                   _state.value = it
+               }
+           } catch (e:Exception){
```
crocsandcoffee, line 31: Also I'd suggest using `runCatchingIgnoreCancelled` here instead and probably you will/could map the error case to a different state, and will bubble up the special CancellationException

data/repository/src/commonMain/kotlin/social/androiddev/common/repository/timeline/TimelineRepoModule.kt
```diff
@@ -0,0 +1,152 @@
...
+import social.androiddev.common.timeline.TimelineItem
+import social.androiddev.domain.timeline.FeedType
+import social.androiddev.domain.timeline.HomeTimelineRepository
+import social.androiddev.domain.timeline.model.StatusUI
+
+/**
+ * Koin module containing all koin/bean definitions for
+ * timeline repository delegates.
+ */
+val timelineRepoModule: Module = module {
```
crocsandcoffee, line 34: This file seems to be doing way more than just providing dependencies. The mappers should probably be moved to the right sites where they are needed and then can be easily tested

crocsandcoffee (commented):
This looks awesome! So excited to start seeing actual user data now. 

I left some suggestions, questions, and some minor comments. I think it should be good to move this out of draft now. 👍🏼

data/network/src/commonMain/kotlin/social/androiddev/common/network/MastodonApi.kt
```diff
@@ -77,4 +78,12 @@ interface MastodonApi {
...
      */
     suspend fun getInstance(domain: String? = null): Result<Instance>
+
+    /**
+     * Fetch home feed for a particular user
+     * @see https://docs.joinmastodon.org/methods/timelines/#home
+     * @param accessToken representing the user
+     * @return a list of [Status]
+     */
+    suspend fun getHomeFeed(domain: String, accessToken: String): Result<List<Status>>
```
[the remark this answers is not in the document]
digitalbuddha: it returns 20 entries by default. Figured that's enough to get the UI up and running and then we can add the paging

data/network/src/commonMain/kotlin/social/androiddev/common/network/MastodonApiKtor.kt
```diff
@@ -124,4 +125,13 @@ internal class MastodonApiKtor(
             Result.failure(exception = exception)
         }
     }
+
+    override suspend fun getHomeFeed(domain: String, accessToken: String): Result<List<Status>> {
+        return runCatchingIgnoreCancelled<List<Status>> {
+           httpClient.get("https://$domain/api/v1/timelines/home") {
```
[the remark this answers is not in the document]
digitalbuddha: done

data/network/src/commonMain/kotlin/social/androiddev/common/network/model/Application.kt
```diff
@@ -17,9 +17,8 @@ import kotlinx.serialization.Serializable
  */
 @Serializable
 data class Application(
-    val id: String,
```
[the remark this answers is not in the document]
digitalbuddha: The changes I made were due to crashes when reading top 20 of my user of androiddev.social. How do we figure out what actually exists/nullability rules?

data/network/src/commonMain/kotlin/social/androiddev/common/network/MastodonApi.kt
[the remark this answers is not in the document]
crocsandcoffee: Perfect. Works for now. 👍

domain/timeline/src/commonMain/kotlin/social/androiddev/domain/timeline/model/StatusUI.kt
digitalbuddha: naming, fixed to StatusLocal to denote it being the local representation

ui/timeline/src/commonMain/kotlin/social/androiddev/timeline/TootContent.kt
digitalbuddha: we're going to have to format the html. I figured this allows us to split the message by html token, format and put back together. I'd use AndroidView and html builder if this was android only, not sure what to do on other platforms if we don't hand parse

ui/timeline/src/commonMain/kotlin/social/androiddev/timeline/navigation/TimelineViewModel.kt
digitalbuddha: done

ui/timeline/src/commonMain/kotlin/social/androiddev/timeline/navigation/TimelineViewModel.kt
digitalbuddha: removed entirely, store returns data/error/loading now without ever throwing an exception

data/repository/src/commonMain/kotlin/social/androiddev/common/repository/timeline/TimelineRepoModule.kt
digitalbuddha: refactored into delegates

digitalbuddha:
UserAvatar does not always load, I think we need to use coil or something

data/persistence/src/androidMain/kotlin/social/androiddev/common/persistence/di/AndroidPersistenceModule.kt
digitalbuddha: yup makes sense, will make separate tables when building the next feed

data/repository/build.gradle.kts
digitalbuddha: yup yup will add to catalog. It's a kotlin issue where android gets wrong version and gets a class def not found https://bytemeta.vip/repo/Kotlin/kotlinx.atomicfu/issues/145

data/network/build.gradle.kts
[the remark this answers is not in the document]
digitalbuddha: I'll do this in a follow up

ui/timeline/src/commonMain/kotlin/social/androiddev/timeline/navigation/TimelineComponent.kt
digitalbuddha: did it as ``` val state: StateFlow<StoreResponse<List<FeedItemState>>> = homeTimelineRepository
        .read(FeedType.Home, refresh = true)
        .mapLatest(::render)
        .stateIn(scope, SharingStarted.Eagerly, StoreResponse.Loading(ResponseOrigin.Cache))

    private fun render(it: StoreResponse<List<StatusLocal>>): StoreResponse.Data<List<FeedItemState>> {
        return when (val response: StoreResponse<List<StatusLocal>> = it) {
            is StoreResponse.Data -> {
                val result = StoreResponse.Data(
                    response.value.map {
                        FeedItemState(
                            id = it.remoteId,
                            userAvatarUrl = it.avatarUrl,
                            date = it.createdAt,
                            username = it.userName,
                            acctAddress = it.accountAddress,
                            message = it.content,
                            images = emptyList(),
                            videoUrl = null,
                        )
                    },
                    response.origin
                )
                result
            }

            else -> {
                StoreResponse.Data(emptyList(), it.origin)
            }
        }
    }
```

**Comment by digitalbuddha** [comment]
All comments addressed - Note - I could not write tests for the TimelineSourceOfTruth due to not being able to access DB in that layer. I'll do an interface around it in follow up (when adding db writes logic)

File: ui/timeline/src/commonMain/kotlin/social/androiddev/timeline/navigation/TimelineViewModel.kt [hunk 55]
```diff
@@ -0,0 +1,73 @@
...
+    private val scope = CoroutineScope(mainContext + SupervisorJob())
+    private val _state =
+        MutableStateFlow<StoreResponse<List<FeedItemState>>>(StoreResponse.Loading(ResponseOrigin.SourceOfTruth))
+    val state: StateFlow<StoreResponse<List<FeedItemState>>> = homeTimelineRepository
+        .read(FeedType.Home, refresh = true)
+        .mapLatest(::render)
+        .stateIn(scope, SharingStarted.Eagerly, StoreResponse.Loading(ResponseOrigin.Cache))
+
+    private fun render(it: StoreResponse<List<StatusLocal>>): StoreResponse.Data<List<FeedItemState>> {
+        return when (val response: StoreResponse<List<StatusLocal>> = it) {
```
crocsandcoffee, line 44:
```suggestion
    private fun render(response: StoreResponse<List<StatusLocal>>): StoreResponse.Data<List<FeedItemState>> {
        return when (response) {
```

domain/timeline/src/commonMain/kotlin/social/androiddev/domain/timeline/HomeTimelineRepository.kt
```diff
@@ -0,0 +1,25 @@
...
+
+interface HomeTimelineRepository {
+    fun read(
+        feedType: FeedType,
+        refresh: Boolean = false
+    ): Flow<StoreResponse<List<StatusLocal>>>
+}
+
+sealed class FeedType(val type: String) {
+    object Home : FeedType("HOME")
```
crocsandcoffee, line 24: I think the domain model FeedType could just be an Enum. And you can map this Enum -> JSON string that needs to be passed to the API in the RealHomeTimelineRepostory.

Just a thought, I don't feel strongly.

crocsandcoffee (approved):
Whooohooooo! Excited to log in and view my own home timeline 🌟

data/network/src/commonMain/kotlin/social/androiddev/common/network/model/Status.kt
```diff
@@ -21,13 +21,13 @@ data class Status(
...
     @SerialName("created_at") val createdAt: String,
-    @SerialName("account") val account: Account,
+    @SerialName("account") val account: Account? = null,
     @SerialName("content") val content: String,
     @SerialName("visibility") val visibility: Privacy,
     @SerialName("sensitive") val sensitive: Boolean,
     @SerialName("spoiler_text") val spoilerText: String,
-    @SerialName("media_attachments") val mediaAttachments: List<Attachment>,
-    @SerialName("application") val application: Application,
+    @SerialName("media_attachments") val mediaAttachments: List<Attachment>? = emptyList(),
```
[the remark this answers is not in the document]
crocsandcoffee: The reason I am suggesting defaulting to null is that it seems safer in the event there are no media attachments. I don't think we will ever get an empty list. Either we get a list back or null, if no data. (Of course, we can still check for empty higher up in the chain)

data/repository/src/commonMain/kotlin/social/androiddev/common/repository/timeline/TimelineFetcher.kt
```diff
@@ -0,0 +1,26 @@
...
+
+fun MastodonApi.timelineFetcher(authStorage: DodoAuthStorage): Fetcher<FeedType, List<StatusDB>> =
+    Fetcher.of { key: FeedType ->
+        when (key) {
+            is FeedType.Home -> {
+                getHomeFeed(
+                    authStorage.currentDomain!!,
+                    authStorage.getAccessToken(authStorage.currentDomain!!)!!
+                )
+                    .getOrThrow()
```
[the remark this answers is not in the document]
digitalbuddha: yessir any exception thrown into a store gets emitted as a StoreResponse.Error.Message

domain/timeline/src/commonMain/kotlin/social/androiddev/domain/timeline/model/StatusLocal.kt
```diff
@@ -9,15 +9,21 @@
...
+    val repliesCount: Long = 0,
+    val reblogsCount: Long = 0,
+    val favoritesCount: Long = 0,
     val content: String,
-    val account: Account,
+    val account: Account? = null,
     val sensitive: Boolean = false,
     val spoilerText: String? = null,
-    val visibility: Visibility = Visibility.PUBLIC
+    val visibility: String = "Public",
```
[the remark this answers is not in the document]
digitalbuddha: changes back was debugging

domain/timeline/src/commonMain/kotlin/social/androiddev/domain/timeline/HomeTimelineRepository.kt
digitalbuddha: let me do as follow up, I had trouble mapping the enum mostly from not working with enums for too long :-)

PR #103, opened 2022-12-18
Gradle improvements

evant:
## 📑 What does this PR do?
- Extract common gradle config into convention plugins
- Update AGP to 7.3.1
- update gradle to 7.6

# ✅ Checklist

Both the desktop & android apps where run to ensure they still built correctly


app-android/build.gradle.kts
```diff
@@ -3,20 +3,16 @@ plugins {
...
 android {
     namespace = "social.androiddev.dodo"
-    compileSdk = compileSDKVersion
+
+    compileSdk = libs.versions.android.compileSdk.get().toInt()
 
     defaultConfig {
-        minSdk = minSDKVersion
-        targetSdk = targetSDKVersion
+        minSdk = libs.versions.android.minSdk.get().toInt()
```
evant, line 14: Slightly uglier access but it keeps the version info in libs.versions.toml like other versions, what do you think?

build-logic/convention-plugins/build.gradle.kts
```diff
@@ -2,15 +2,11 @@ plugins {
     `kotlin-dsl`
 }
 
-gradlePlugin {
```
evant, line 5: Rolled the code-quality config into the common convention plugin, since we should apply it to all modules anyway. Could keep it separate if there's a huge desire to though.

data/network/build.gradle.kts
```diff
@@ -53,64 +23,44 @@ kotlin {
...
+        val androidMain by getting {
             dependencies {
                 implementation(libs.io.ktor.client.cio)
             }
         }
 
-
         // desktop
-        getByName("desktopMain") {
+        val desktopMain by getting {
```
evant, line 34: Changed access everywhere to the `by getting` syntax to be consistent.

data/persistence/src/androidMain/AndroidManifest.xml
```diff
@@ -1,3 +1,2 @@
 <?xml version="1.0" encoding="utf-8"?>
-<manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="social.androiddev.data.persistence" />
+<manifest xmlns:android="http://schemas.android.com/apk/res/android" />
```
evant, line 2: Was already defined in the gradle file.

gradle.properties
```diff
@@ -37,4 +37,5 @@ kotlin.code.style=official
 android.suppressUnsupportedCompileSdk=33
 
 # suppress Kotlin Multiplatform Projects are an "Alpha feature" warning
-kotlin.mpp.stability.nowarn=true
\ No newline at end of file
+kotlin.mpp.stability.nowarn=true
+kotlin.native.ignoreDisabledTargets=true
```
evant, line 41: Removes warnings about not being able to build iOS on linux, don't have strong opinions about it.

build-logic/settings.gradle.kts
```diff
@@ -3,8 +3,9 @@ rootProject.name = "build-logic"
 dependencyResolutionManagement {
     repositories {
         google()
-        gradlePluginPortal()
         mavenCentral()
+        gradlePluginPortal()
```
crocsandcoffee, line 7: I got this similar feedback on a PR of mine recently which I agree with. The ordering should be:
gradlePluginPortal()
mavenCentral()
Google()

And we actually can remove jerbrain compose dev url, it's available via mavenCentral() already

data/network/build.gradle.kts
crocsandcoffee: Yes thanks for doing this. It was bugging me that we were using different conventions across all the build files. Although I prefer named() or getByName() but I don't feel strongly. At least now it's consistent

data/persistence/src/androidMain/AndroidManifest.xml
crocsandcoffee: Good catch - we shouldn't be setting this in manifest anymore regardless

gradle.properties
crocsandcoffee: I'm fine with this - but would prefer a comment stating this with a relevant link if there is one

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.5.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.6-bin.zip
+networkTimeout=10000
```
crocsandcoffee, line 4: Was network timeout added by you?

app-android/build.gradle.kts
crocsandcoffee: I won't lose sleep over it lol but how about updating the ext in project gradle to still have minSDK, etc but instead of hardcoded values, it references these versions from the catalog? At least the ugliness will be contained to a single place that way. Don't feel strongly either way.

build-logic/convention-plugins/src/main/kotlin/social.androiddev.library.common.gradle.kts
```diff
@@ -0,0 +1,62 @@
+plugins {
```
crocsandcoffee, line 1: I like the idea of having a top level comment on these different plugins with a description of what they are for and/or do. Mind adding that to the three you created?

build-logic/convention-plugins/src/main/kotlin/social.androiddev.library.gradle.kts
```diff
@@ -0,0 +1,65 @@
+plugins {
+    id("social.androiddev.library.common")
+}
+
+val libs = the<org.gradle.accessors.dm.LibrariesForLibs>()
+
+kotlin {
+    jvm("desktop")
```
crocsandcoffee, line 8: Please remove "desktop". I believe it actually doesn't do anything and is there just for assigning a name

build-logic/convention-plugins/src/main/kotlin/social.androiddev.library.gradle.kts
```diff
@@ -0,0 +1,65 @@
...
+                implementation(libs.org.jetbrains.kotlin.test.junit)
+            }
+        }
+        val desktopTest by getting {
+            dependencies {
+                implementation(kotlin("test"))
+                implementation(libs.org.jetbrains.kotlin.test.junit)
+            }
+        }
+        val iosX64Test by getting
```
crocsandcoffee, line 54: I *_think_* you can remove all of these targets that have empty blocks . They're redundant and not needed. Keeps the file cleaner/less polluted that way

build-logic/convention-plugins/src/main/kotlin/social.androiddev.library.ui.gradle.kts
```diff
@@ -0,0 +1,38 @@
+plugins {
+    id("social.androiddev.library.common")
+    id("org.jetbrains.compose")
+}
+
+val libs = the<org.gradle.accessors.dm.LibrariesForLibs>()
+
+kotlin {
+    jvm("desktop")
```
crocsandcoffee, line 9: Same comment. Please update to just `jvm()`

crocsandcoffee (commented):
Looks good. Thanks for this cleanup!!!

gradle/wrapper/gradle-wrapper.properties
evant: Updating the gradle wrapper added it

build-logic/convention-plugins/build.gradle.kts
```diff
@@ -2,15 +2,11 @@ plugins {
...
-    }
-}
-
 dependencies {
     implementation(libs.com.diffplug.spotless.gradle.plugin)
+    implementation(libs.org.jetbrains.kotlin.gradle.plugin)
+    implementation(libs.org.jetbrains.compose.gradle.plugin)
+    implementation(libs.com.android.tools.build.gradle)
+    // hack to access version catalogue https://github.com/gradle/gradle/issues/15383
+    implementation(files(libs.javaClass.superclass.protectionDomain.codeSource.location))
```
krizzu, line 11: What about creating a version catalog in `settings.gradle.kts` - would it work instead of the hack?

```
    versionCatalogs {
        create("libs") {
            from(files(File(rootDir, "../gradle/libs.versions.toml")))
        }
    }
```

build-logic/convention-plugins/build.gradle.kts
krizzu: It seems we lost `spotless` from `app-*` modules, so maybe we could have another plugin named `.app` where we apply it there?

krizzu (commented):
I really like the improvements 👍

build-logic/convention-plugins/build.gradle.kts
evant: That is already in settings.gradle.kts

app-android/build.gradle.kts
evant: It's only used in 2 places now so kinda meh about it, honestly might be better to figure out a way to just define it once in the convention plugins.

data/network/build.gradle.kts
evant: Don't feel strongly either, but the one benifit for the above syntax is you can't typo it doing:

```
val iosMain by creating {
     dependsOn(commonMain)
}
```

crocsandcoffee: Yup that's fine. As long as we stick with one moving forward than works for me 👍

build-logic/convention-plugins/build.gradle.kts
evant: Fixed

gradle.properties
```diff
@@ -17,7 +17,7 @@ org.gradle.jvmargs=-Xmx7168m -XX:MaxMetaspaceSize=2g
 org.gradle.parallel=true
 # Default is number of CPU processors.
 # org.gradle.workers.max=
-
+se
```
crocsandcoffee, line 20: Typo?

crocsandcoffee (approved):
I dig it. Just one comment remaining from me which I think is just a typo in one of the build scripts. 

Nice work 👏👏👏

crocsandcoffee:
Seems CI is failing. Please take a quick peep and resolve any errors before we merge. I can also help take a look tomorrow when I get some time. 

You should be able to also run `gradlew check` locally and everything should pass.

evant:
1.  Dependency 'androidx.activity:activity-compose:1.6.1' requires libraries and applications that
           depend on it to compile against version 33 or later of the
           Android APIs.

                :app-android is currently compiled against android-30.

           Recommended action: Update this project to use a newer compileSdkVersion
           of at least 33, for example 33.
          
ok now I'm confused how this was working before

build-logic/convention-plugins/build.gradle.kts
krizzu: I missed that 🙈

build-logic/convention-plugins/src/main/kotlin/social.androiddev.library.gradle.kts
```diff
@@ -7,26 +10,12 @@ val libs = the<org.gradle.accessors.dm.LibrariesForLibs>()
 kotlin {
     jvm("desktop")
     android()
-    iosX64()
-    iosArm64()
-    iosSimulatorArm64()
+    ios()
```
krizzu, line 13: does ios target accumulate all others (x64, arm, simulators)? If so, that's awesome

krizzu (approved):
Awesome 👏

build-logic/convention-plugins/src/main/kotlin/social.androiddev.library.gradle.kts
evant: yep!

PR #109, opened 2022-12-21
Miken/feed UI

digitalbuddha:
## 📑 What does this PR do?
Adds an image loading library
Adds a html parsing library
displays the avatar 
parses content html and displays as annotated string with links in blue

# ✅ Checklist

Screen Shot 2022-12-21 at 9 53 25 AM

### DIFF
--- a/app-android/build.gradle.kts
+++ b/app-android/build.gradle.kts
@@ -26,6 +26,13 @@ android {
         targetCompatibility = JavaVersion.VERSION_11
     }
 
+    packagingOptions {
+        exclude ("META-INF/DEPENDENCIES")
+        exclude ("META-INF/NOTICE")
+        exclude ("mozilla/public-suffix-list.txt")
+
+    }
+
     kotlinOptions {
         jvmTarget = "11"
     }
--- a/ui/common/build.gradle.kts
+++ b/ui/common/build.gradle.kts
@@ -66,6 +66,15 @@ kotlin {
                 implementation(compose.desktop.common)
             }
         }
+        named("commonTest") {
+            dependencies {
+                implementation(kotlin("test"))
+                implementation(libs.org.jetbrains.kotlin.test.common)
+                implementation(libs.org.jetbrains.kotlin.test.annotations.common)
+                implementation(libs.org.jetbrains.kotlinx.coroutines.test)
+            }
+        }
+
     }
 
     tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile> {
--- a/ui/common/build.gradle.kts
+++ b/ui/common/build.gradle.kts
@@ -18,6 +18,9 @@ android {
         minSdk = minSDKVersion
         targetSdk = targetSDKVersion
     }
+    packagingOptions {
+        exclude ("META-INF/DEPENDENCIES")
+    }
 
     compileOptions {
         sourceCompatibility = JavaVersion.VERSION_11
@@ -45,6 +48,9 @@ kotlin {
                 implementation(compose.material)
                 api(libs.com.arkivanov.decompose)
                 api(libs.com.arkivanov.decompose.extensions.compose.jetbrains)
+                implementation("com.alialbaali.kamel:kamel-image:0.4.0")
+                implementation("it.skrape:skrapeit:1.2.2")
+
             }
         }
 
--- a/ui/common/src/commonMain/kotlin/social/androiddev/common/composables/UserAvatar.kt
+++ b/ui/common/src/commonMain/kotlin/social/androiddev/common/composables/UserAvatar.kt
@@ -9,6 +9,7 @@
  */
 package social.androiddev.common.composables
 
+import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.width
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.runtime.Composable
@@ -26,10 +27,9 @@ fun UserAvatar(
     url: String
 ) {
     AsyncImage(
-        load = { loadImageIntoPainter(url = url) },
-        painterFor = { remember { it } },
+        url = url,
         contentDescription = "User avatar",
-        modifier = modifier.width(48.dp).clip(RoundedCornerShape(5.dp))
+        modifier = modifier.width(48.dp).height(48.dp).clip(RoundedCornerShape(5.dp))
     )
 }
 
--- a/ui/common/src/commonMain/kotlin/social/androiddev/common/utils/AsyncImage.kt
+++ b/ui/common/src/commonMain/kotlin/social/androiddev/common/utils/AsyncImage.kt
@@ -9,56 +9,29 @@
  */
 package social.androiddev.common.utils
 
-import androidx.compose.foundation.Image
-import androidx.compose.foundation.layout.Spacer
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.getValue
-import androidx.compose.runtime.produceState
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
-import androidx.compose.ui.graphics.painter.Painter
 import androidx.compose.ui.layout.ContentScale
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.withContext
-import social.androiddev.common.network.util.runCatchingIgnoreCancelled
+import io.kamel.image.KamelImage
+import io.kamel.image.lazyPainterResource
 
 /**
  * Use this helper until we switch to a image loading library which supports multiplatform
  */
 @Composable
-fun <T> AsyncImage(
+fun AsyncImage(
     contentDescription: String,
     modifier: Modifier = Modifier,
     alignment: Alignment = Alignment.Center,
     contentScale: ContentScale = ContentScale.Fit,
-    load: suspend () -> T,
-    painterFor: @Composable (T) -> Painter,
+    url:String
 ) {
-    val image: T? by produceState<T?>(null) {
-        value = withContext(Dispatchers.IO) {
-            runCatchingIgnoreCancelled {
-                load()
-            }.fold(
-                onSuccess = { it },
-                onFailure = { t ->
-                    t.printStackTrace()
-                    null
-                }
-            )
-        }
-    }
-
-    if (image != null) {
-        Image(
-            painter = painterFor(image!!),
-            contentDescription = contentDescription,
-            contentScale = contentScale,
-            modifier = modifier,
-            alignment = alignment,
-        )
-    } else {
-        Spacer(
-            modifier = modifier
-        )
-    }
+    KamelImage(
+        resource = lazyPainterResource(data = url),
+        contentDescription =contentDescription,
+        modifier = modifier,
+        alignment = alignment,
+        contentScale = contentScale
+    )
 }
--- a/ui/common/src/commonMain/kotlin/social/androiddev/common/utils/HtmlParser.kt
+++ b/ui/common/src/commonMain/kotlin/social/androiddev/common/utils/HtmlParser.kt
@@ -1,111 +1,66 @@
 package social.androiddev.common.utils
 
-import androidx.compose.foundation.layout.padding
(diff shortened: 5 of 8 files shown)


digitalbuddha:
this is stacked on my other diff

ui/signed-out/src/commonMain/kotlin/social/androiddev/signedout/landing/LandingContent.kt
```diff
@@ -73,12 +71,12 @@ fun LandingContent(
     onGetStartedClicked: () -> Unit,
     appIcon: @Composable () -> Unit = {
         AsyncImage(
-            load = { loadImageIntoPainter(url = "https://via.placeholder.com/200x200/6FA4DE/010041?text=Dodo") },
-            painterFor = { remember { it } },
+            url = "https://via.placeholder.com/200x200/6FA4DE/010041?text=Dodo",
             contentDescription = "App Logo",
             modifier = Modifier
                 .padding(horizontal = 48.dp)
-                .size(240.dp)
```
crocsandcoffee, line 81: size already sets the width + height to the same value. Or is there a different reason for breaking this out?

ui/common/src/commonMain/kotlin/social/androiddev/common/utils/HtmlParser.kt
```diff
@@ -0,0 +1,77 @@
...
+                withStyle(style = SpanStyle(color = Color.Blue, fontWeight = FontWeight.Bold)) {
+                    appendLink(key, value)
+                }
+                restOfPlainText = restOfPlainText.substringAfter(key)
+            }
+            append(restOfPlainText)
+        }
+
+    }
+   return  buildAnnotatedString {  }
```
crocsandcoffee, line 63: I think it would be easier to read to just have the early guard at the top of this function, where you just return right away if <p> is missing

ui/common/src/commonMain/kotlin/social/androiddev/common/utils/HtmlParser.kt
```diff
@@ -0,0 +1,77 @@
...
+}
+
+
+fun AnnotatedString.Builder.appendLink(linkText: String, linkUrl: String) {
```
crocsandcoffee, line 67: This and onLinkClick below should probably be private

ui/common/src/commonMain/kotlin/social/androiddev/common/utils/HtmlParser.kt
```diff
@@ -0,0 +1,77 @@
...
+import it.skrape.selects.eachText
+import it.skrape.selects.html5.a
+import it.skrape.selects.html5.body
+
+
+/**
+ * Takes a String reciever of html text
+ * converts it to an annotated string of text and links
+ */
+fun String.renderHtml(): AnnotatedString {
```
crocsandcoffee, line 20: I'm wondering if this should instead be moved to the VM? The TootContent UI model could have AnnotatedString in place of String for the message field.

ui/common/src/commonMain/kotlin/social/androiddev/common/utils/AsyncImage.kt
```diff
@@ -9,56 +9,29 @@
...
 @Composable
-fun <T> AsyncImage(
+fun AsyncImage(
     contentDescription: String,
     modifier: Modifier = Modifier,
     alignment: Alignment = Alignment.Center,
     contentScale: ContentScale = ContentScale.Fit,
-    load: suspend () -> T,
-    painterFor: @Composable (T) -> Painter,
+    url:String
```
crocsandcoffee, line 28: Similar to how coil for compose works: it would be nice to never have to differentiate at call site if we should use AsyncImage or built in Image, and be able to use a single Image compose wrapper for rendering url or a drawable. 

Can we change this to `Any` if KamelImage supports it?

ui/common/src/commonMain/kotlin/social/androiddev/common/utils/AsyncImage.kt
```diff
@@ -9,56 +9,29 @@
...
-            contentScale = contentScale,
-            modifier = modifier,
-            alignment = alignment,
-        )
-    } else {
-        Spacer(
-            modifier = modifier
-        )
-    }
+    KamelImage(
```
crocsandcoffee, line 30: Is this truly KMP support or just JVM?

ui/common/src/commonMain/kotlin/social/androiddev/common/composables/UserAvatar.kt
```diff
@@ -26,10 +27,9 @@ fun UserAvatar(
     url: String
 ) {
     AsyncImage(
-        load = { loadImageIntoPainter(url = url) },
-        painterFor = { remember { it } },
+        url = url,
         contentDescription = "User avatar",
-        modifier = modifier.width(48.dp).clip(RoundedCornerShape(5.dp))
+        modifier = modifier.width(48.dp).height(48.dp).clip(RoundedCornerShape(5.dp))
```
crocsandcoffee, line 32: modifier.size(48.dp) should already set it for both dimensions

ui/common/build.gradle.kts
```diff
@@ -45,6 +48,9 @@ kotlin {
                 implementation(compose.material)
                 api(libs.com.arkivanov.decompose)
                 api(libs.com.arkivanov.decompose.extensions.compose.jetbrains)
+                implementation("com.alialbaali.kamel:kamel-image:0.4.0")
+                implementation("it.skrape:skrapeit:1.2.2")
```
crocsandcoffee, line 52: Is this also JVM only? Maybe we can use this as a resource and roll our own purely kotlin parser which also supports iOS

app-android/build.gradle.kts
```diff
@@ -26,6 +26,13 @@ android {
         targetCompatibility = JavaVersion.VERSION_11
     }
 
+    packagingOptions {
+        exclude ("META-INF/DEPENDENCIES")
```
crocsandcoffee, line 30: I've always wondering what this stuff is for? Is this generated or you had to add manually to fix a build issue?

crocsandcoffee (commented):
Just did a quick early review for it..

This looks really good! I think you should be able to remove the platform specific logic for loading urls now, I didn't see that code delete in this PR.

app-android/build.gradle.kts
digitalbuddha: Manually to fix a build issue. My naive understanding is that multiple libraries have the same metadata file name and clash

ui/common/src/commonMain/kotlin/social/androiddev/common/utils/HtmlParser.kt
digitalbuddha: good idea

ui/common/src/commonMain/kotlin/social/androiddev/common/utils/AsyncImage.kt
digitalbuddha: yup yup it seems to delegate to same chain

ui/common/src/commonMain/kotlin/social/androiddev/common/composables/UserAvatar.kt
digitalbuddha: for some reason it was not rendering the avatars after switching to the new library. We can play with it later but yeah it never assumed a height

ui/common/build.gradle.kts
digitalbuddha: yup yup jvm only. I think this gets us off the ground and agree we would need something more robust in future

ui/common/src/commonMain/kotlin/social/androiddev/common/utils/HtmlParser.kt
```diff
@@ -0,0 +1,84 @@
...
+    val newlineReplace = this.replace("<br>", "\n")
+
+    if (newlineReplace.contains("<p>").not()) return buildAnnotatedString { }
+
+    val content = htmlDocument(newlineReplace)
+
+    val body = content.body {
```
digitalbuddha, line 35: At this point, the only thing we are using skrape for is to get the paragraph body without any other tags and then for getting the list of links.

We can hand roll that logic ourselves and target ios

ui/common/src/commonMain/kotlin/social/androiddev/common/utils/AsyncImage.kt
digitalbuddha: just jvm, but ios has native support afaik

data/repository/src/commonTest/kotlin/social/androiddev/common/repository/timeline/RealHomeTimelineRepositoryTest.kt
```diff
@@ -28,8 +28,7 @@ import kotlin.test.assertTrue
 import kotlin.test.fail
 
 class RealHomeTimelineRepositoryTest {
-    @Test
-    fun sucessTest(): TestResult {
+    @Test fun sucessTest(): TestResult {
```
crocsandcoffee, line 31:
```suggestion
    @Test fun successTest(): TestResult {
```

ui/common/build.gradle.kts
```diff
@@ -14,6 +17,22 @@ kotlin {
...
+
+            }
+        }
+        val commonTest by getting {
+            dependencies {
+                implementation(kotlin("test"))
+                implementation(libs.org.jetbrains.kotlin.test.common)
+                implementation(libs.org.jetbrains.kotlin.test.annotations.common)
```
crocsandcoffee, line 29: Formatting here looks off, but it could be because I'm on my phone

ui/common/src/commonMain/kotlin/social/androiddev/common/utils/AsyncImage.kt
crocsandcoffee: Cool, we should follow-up on this and change this to an expect/actual

ui/common/src/commonMain/kotlin/social/androiddev/common/utils/HtmlParser.kt
```diff
@@ -0,0 +1,83 @@
...
+import it.skrape.selects.eachLink
+import it.skrape.selects.eachText
+import it.skrape.selects.html5.a
+import it.skrape.selects.html5.body
+
+/**
+ * Takes a String reciever of html text
+ * converts it to an annotated string of text and links
+ */
+fun String.renderHtml(): AnnotatedString {
```
crocsandcoffee, line 28: This probably should be either moved to the ViewModel itself and not a public API, or renamed to something that is more descriptive. something like 

```
fun String.extractTootContentFromHtml()
```

crocsandcoffee (approved):
🎉

ui/common/build.gradle.kts
```diff
@@ -14,6 +17,22 @@ kotlin {
                 implementation(projects.data.network)
                 api(libs.com.arkivanov.decompose)
                 api(libs.com.arkivanov.decompose.extensions.compose.jetbrains)
+                implementation("com.alialbaali.kamel:kamel-image:0.4.0")
+                implementation("it.skrape:skrapeit:1.2.2")
```
mboukadir, line 21: I think the declaration of dependencies is missing in the `libs.versions.tom` file.

ui/common/build.gradle.kts
digitalbuddha: ran spotless apply just in case

ui/common/src/commonMain/kotlin/social/androiddev/common/utils/HtmlParser.kt
digitalbuddha: renamed to `extractContentFromMicroFormat` 
When looking at Tusky, I noticed there are a few places where we get this type of text (profile and server about is another)

Going to leave as a utility for now

ui/common/build.gradle.kts
digitalbuddha: thanks fixed

Issue #107, opened 2022-12-19
[FEAT]: Settings and Logout 


krizzu:
### Description

>  for now, is just add a dummy TopAppBar to the root screen hierarchy when logged in, and add a settings icon that takes you to a new screen with a simple logout button. You'll need to also add logic for clearing/deleting the right user session data which may only be local, not sure yet. Plz feel free to create a GitHub issue with this info if there isn't one already

PR #111, opened 2022-12-21
feat(logout): Keep user signed in

krizzu:
## 📑 What does this PR do?

As part of https://github.com/AndroidDev-social/DodoForMastodon/issues/107, navigate user to sign in flow once authenticated. 

For now, the auth status does not redirect to Timeline (waiting for #107), otherwise we might get stuck.

Things done:
- Desktop DB implementation saves data to disk, instead of memory
- Added AuthState model, to reflect the authorization state for user
- Moved Root to `root` module, moved Splash to `splash` module
- Added [Turbine](https://github.com/cashapp/turbine) for testing flows

# ✅ Checklist

- Use case to control auth state, exposed via flow

### DIFF
--- a/app-desktop/src/jvmMain/kotlin/social/androiddev/desktop/Main.kt
+++ b/app-desktop/src/jvmMain/kotlin/social/androiddev/desktop/Main.kt
@@ -24,8 +24,8 @@ import kotlinx.coroutines.Dispatchers
 import org.koin.core.context.startKoin
 import social.androiddev.common.di.appModule
 import social.androiddev.common.theme.DodoTheme
-import social.androiddev.root.composables.RootContent
-import social.androiddev.root.navigation.DefaultRootComponent
+import social.androiddev.root.root.DefaultRootComponent
+import social.androiddev.root.root.RootContent
 
 fun main() {
     startKoin {
--- a/data/persistence/src/desktopMain/kotlin/social/androiddev/common/persistence/di/DesktopPersistenceModule.kt
+++ b/data/persistence/src/desktopMain/kotlin/social/androiddev/common/persistence/di/DesktopPersistenceModule.kt
@@ -21,6 +21,7 @@ import social.androiddev.common.persistence.AuthenticationDatabase
 import social.androiddev.common.persistence.localstorage.DodoAuthStorage
 import social.androiddev.common.persistence.localstorage.DodoAuthStorageImpl
 import social.androiddev.common.timeline.TimelineDatabase
+import java.io.File
 
 /**
  * Koin DI module for all desktop specific persistence dependencies
@@ -39,7 +40,13 @@ actual val persistenceModule: Module = module {
     }
 
     single {
-        val driver = JdbcSqliteDriver(url = JdbcSqliteDriver.IN_MEMORY).also { driver ->
+        // refers to the user directory, such us /Users/my_username or /home/my_username or C:\Users\my_username
+        val dbDir = File(System.getProperty("user.home"), "dodo")
+        if (!dbDir.exists()) {
+            dbDir.mkdirs()
+        }
+        val dbFile = File(dbDir, AUTH_DB_NAME)
+        val driver = JdbcSqliteDriver(url = "jdbc:sqlite:${dbFile.absolutePath}").also { driver ->
             AuthenticationDatabase.Schema.create(driver = driver)
         }
         AuthenticationDatabase(driver)
--- a/domain/authentication/src/commonMain/kotlin/social/androiddev/domain/authentication/repository/AuthenticationRepository.kt
+++ b/domain/authentication/src/commonMain/kotlin/social/androiddev/domain/authentication/repository/AuthenticationRepository.kt
@@ -12,6 +12,7 @@
  */
 package social.androiddev.domain.authentication.repository
 
+import kotlinx.coroutines.flow.Flow
 import social.androiddev.domain.authentication.model.ApplicationOAuthToken
 import social.androiddev.domain.authentication.model.NewAppOAuthToken
 
@@ -22,7 +23,7 @@ interface AuthenticationRepository {
         clientName: String,
         redirectUris: String,
         scopes: String,
-        website: String?
+        website: String?,
     ): NewAppOAuthToken?
 
     suspend fun saveApplication(
@@ -42,4 +43,6 @@ interface AuthenticationRepository {
     val selectedServer: String?
 
     suspend fun getApplicationOAuthToken(server: String): ApplicationOAuthToken?
+
+    suspend fun getIsAccessTokenPresent(): Flow<Boolean?>
 }
--- a/domain/authentication/src/commonMain/kotlin/social/androiddev/domain/authentication/repository/AuthenticationRepository.kt
+++ b/domain/authentication/src/commonMain/kotlin/social/androiddev/domain/authentication/repository/AuthenticationRepository.kt
@@ -44,5 +44,5 @@ interface AuthenticationRepository {
 
     suspend fun getApplicationOAuthToken(server: String): ApplicationOAuthToken?
 
-    suspend fun getIsAccessTokenPresent(): Flow<Boolean?>
+    suspend fun getIsAccessTokenPresent(): Flow<Boolean>
 }
--- a/domain/authentication/src/commonMain/kotlin/social/androiddev/domain/authentication/usecase/GetAuthStatus.kt
+++ b/domain/authentication/src/commonMain/kotlin/social/androiddev/domain/authentication/usecase/GetAuthStatus.kt
@@ -10,27 +10,25 @@
  * You should have received a copy of the GNU General Public License along with Dodo.
  * If not, see <https://www.gnu.org/licenses/>.
  */
-package social.androiddev.common.repository.di
+package social.androiddev.domain.authentication.usecase
 
-import kotlinx.coroutines.Dispatchers
-import org.koin.core.module.Module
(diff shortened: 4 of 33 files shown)


ui/root/src/commonMain/kotlin/social/androiddev/root/composables/RootContent.kt
```diff
@@ -19,7 +19,8 @@ import com.arkivanov.decompose.ExperimentalDecomposeApi
 import com.arkivanov.decompose.extensions.compose.jetbrains.stack.Children
 import com.arkivanov.decompose.extensions.compose.jetbrains.subscribeAsState
 import social.androiddev.root.navigation.RootComponent
-import social.androiddev.root.navigation.SplashComponent
+import social.androiddev.root.splash.SplashComponent
+import social.androiddev.root.splash.SplashContent
```
krizzu, line 23: I extracted `splash` to move related files into one modules

ui/root/src/commonMain/kotlin/social/androiddev/root/splash/DefaultSplashComponent.kt
```diff
@@ -7,15 +7,30 @@
...
 
 class DefaultSplashComponent(
     private val componentContext: ComponentContext,
+    private val mainContext: CoroutineContext,
     private val navigateToLandingInternal: () -> Unit,
     private val navigateToTimelineInternal: () -> Unit,
-) : SplashComponent, ComponentContext by componentContext {
+) : SplashComponent, KoinComponent, ComponentContext by componentContext {
+
+    private val getSelectedAuthToken: GetSelectedApplicationOAuthToken by inject()
```
krizzu, line 27: I'm thinking about a better way to tell if someone is logged in or not - here I check for the existence of auth token. What do you think?

digitalbuddha: I think we should have a state flow of auth state which is exposed from a repository/use case as a flow. This way you can flatmaplatest (I think that's the switchmap equivalent) and be able to react to auth state changes.  For now I think it would be just login and logout but eventually would encode WHICH user is currently viewing the component when we have multi user management.  IMO the beginning of navigation should be mapped from this flow. Change user and you get new components.

krizzu: Yup, agreed - this should be done at Root component level, which then auth state controls which flow will be taken.
I'm going to address this in this PR

data/persistence/src/desktopMain/kotlin/social/androiddev/common/persistence/di/DesktopPersistenceModule.kt
```diff
@@ -39,7 +40,13 @@ actual val persistenceModule: Module = module {
     }
 
     single {
-        val driver = JdbcSqliteDriver(url = JdbcSqliteDriver.IN_MEMORY).also { driver ->
+        // refers to the user directory, such us /Users/my_username or /home/my_username or C:\Users\my_username
+        val dbDir = File(System.getProperty("user.home"), "dodo")
```
krizzu, line 44: This will make sure Desktop platform stores the data, instead keeping it in memory

app-desktop/src/jvmMain/kotlin/social/androiddev/desktop/Main.kt
```diff
@@ -24,8 +24,8 @@ import kotlinx.coroutines.Dispatchers
 import org.koin.core.context.startKoin
 import social.androiddev.common.di.appModule
 import social.androiddev.common.theme.DodoTheme
-import social.androiddev.root.composables.RootContent
-import social.androiddev.root.navigation.DefaultRootComponent
+import social.androiddev.root.root.DefaultRootComponent
+import social.androiddev.root.root.RootContent
```
krizzu, line 28: also moved root content to separate module (for consistency with `splash` module)

ui/root/src/commonMain/kotlin/social/androiddev/root/splash/SplashContent.kt
```diff
@@ -56,12 +59,14 @@ fun SplashContent(
     ) {
         Text("Loading")
 
-        LaunchedEffect(Unit) {
-            // TODO: Hook up to a DI ViewModel
-            if (true) {
+        LaunchedEffect(authStatus) {
+            if (authStatus is AuthStatus.Authorized) {
+                // TODO(krzysztof): Do not navigate to timeline, until logout func is implemented
```
krizzu, line 64: This is because there's no way to navigate back to Welcome screen, if there's a problem fetching the timeline.

thebino (approved):

domain/authentication/src/commonMain/kotlin/social/androiddev/domain/authentication/repository/AuthenticationRepository.kt
```diff
@@ -42,4 +43,6 @@ interface AuthenticationRepository {
     val selectedServer: String?
 
     suspend fun getApplicationOAuthToken(server: String): ApplicationOAuthToken?
+
+    suspend fun getIsAccessTokenPresent(): Flow<Boolean>
```
crocsandcoffee, line 47: NIT - `get` is not needed and can be renamed to `isAccessTokenPresent()`

domain/authentication/src/commonMain/kotlin/social/androiddev/domain/authentication/usecase/GetAuthStatus.kt
```diff
@@ -0,0 +1,30 @@
...
+
+class GetAuthStatus(private val authenticationRepository: AuthenticationRepository) {
+    suspend operator fun invoke(): Flow<AuthStatus> = flow {
+        authenticationRepository.getIsAccessTokenPresent().collect { hasAccessToken ->
+            if (hasAccessToken) {
+                emit(AuthStatus.Authorized)
+            } else {
+                emit(AuthStatus.Unauthorized)
+            }
+        }
```
crocsandcoffee, line 28: instead of collecting a flow here, you can just call `map` on the underlying flow and map it to domain model `AuthStatus`

ui/root/src/commonMain/kotlin/social/androiddev/root/root/RootComponentViewModel.kt
```diff
@@ -0,0 +1,52 @@
...
+import kotlin.coroutines.CoroutineContext
+
+interface RootComponentViewModel : InstanceKeeper.Instance {
+    val authState: StateFlow<AuthStatus?>
+}
+
+fun RootComponentViewModel(coroutineContext: CoroutineContext, getAuthStatus: GetAuthStatus): RootComponentViewModel =
+    RealRootComponentViewModel(coroutineContext, getAuthStatus)
+
+private class RealRootComponentViewModel(coroutineContext: CoroutineContext, getAuthStatus: GetAuthStatus) :
```
crocsandcoffee, line 34: this abstraction is not really needed. You can make the ViewModel internal if you would like, but its really only used via the decompose component, and arguably we could move this logic there and rid away these "view models"

ui/root/src/commonMain/kotlin/social/androiddev/root/root/RootComponentViewModel.kt
```diff
@@ -0,0 +1,52 @@
...
+    private val _authState = MutableStateFlow<AuthStatus?>(null)
+    override val authState = _authState.asStateFlow()
+
+    init {
+        viewModelScope.launch {
+            getAuthStatus().collect { currentStatus ->
+                _authState.value = currentStatus
+            }
+        }
+    }
```
crocsandcoffee, line 47: why not just inline?
```kotlin
    private val _authState = getAuthStatus().stateIn(scope, ...., null)
```

ui/root/src/commonMain/kotlin/social/androiddev/root/splash/SplashContent.kt
```diff
@@ -27,9 +27,11 @@ import social.androiddev.root.navigation.SplashComponent
 @Composable
 fun SplashContent(
     component: SplashComponent,
+    authStatus: AuthStatus?,
```
crocsandcoffee, line 30: the UI/View _should not_ know about or be coupled with `domain` level objects. Please map domain -> UI model if needed

domain/authentication/src/commonMain/kotlin/social/androiddev/domain/authentication/usecase/GetAuthStatus.kt
krizzu: Ah, good call! I had changed that code a few times, so missed that one

ui/root/src/commonMain/kotlin/social/androiddev/root/root/RootComponentViewModel.kt
krizzu: Good call again, thanks! leftover from tinkering 🙈

ui/root/src/commonMain/kotlin/social/androiddev/root/root/RootComponentViewModel.kt
krizzu: Good call re abstraction - leftover again 🙈 
Although, I find our "view models" useful in moving chunks of code away from already overloaded components (where all navigation, configurations, and injections happen).

I don't mind moving this view model logic back to component

PR #124, opened 2023-01-05
feat(logout): Small improvements

krizzu:
## 📑 What does this PR do?

Addressing comments from #111 

# ✅ Checklist

### DIFF
--- a/ui/root/src/commonMain/kotlin/social/androiddev/root/root/RootComponent.kt
+++ b/ui/root/src/commonMain/kotlin/social/androiddev/root/root/RootComponent.kt
@@ -15,7 +15,6 @@ package social.androiddev.root.root
 import com.arkivanov.decompose.router.stack.ChildStack
 import com.arkivanov.decompose.value.Value
 import kotlinx.coroutines.flow.StateFlow
-import social.androiddev.domain.authentication.model.AuthStatus
 import social.androiddev.root.splash.SplashComponent
 import social.androiddev.signedin.navigation.SignedInRootComponent
 import social.androiddev.signedout.root.SignedOutRootComponent
@@ -29,7 +28,7 @@ interface RootComponent {
     val childStack: Value<ChildStack<*, Child>>
 
     // current authorization status
-    val authStatus: StateFlow<AuthStatus?>
+    val authStatus: StateFlow<UiAuthStatus>
 
     /**
      * Supported "Child"s in this navigation stack. These are created from a configuration that
@@ -48,5 +47,11 @@ interface RootComponent {
      */
     sealed interface DeepLink {
         object None : DeepLink
+    }
+
+    sealed class UiAuthStatus {
+        object Unknown : UiAuthStatus()
+        object Authorized : UiAuthStatus()
+        object Unauthorized : UiAuthStatus()
     }
 }
--- a/ui/root/src/commonMain/kotlin/social/androiddev/root/root/RootComponent.kt
+++ b/ui/root/src/commonMain/kotlin/social/androiddev/root/root/RootComponent.kt
@@ -48,10 +48,4 @@ interface RootComponent {
     sealed interface DeepLink {
         object None : DeepLink
     }
-
-    sealed class UiAuthStatus {
-        object Unknown : UiAuthStatus()
-        object Authorized : UiAuthStatus()
-        object Unauthorized : UiAuthStatus()
-    }
 }
--- a/ui/root/src/commonMain/kotlin/social/androiddev/root/root/RootComponentViewModel.kt
+++ b/ui/root/src/commonMain/kotlin/social/androiddev/root/root/RootComponentViewModel.kt
@@ -17,36 +17,35 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.cancel
 import kotlinx.coroutines.flow.MutableStateFlow
-import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.asStateFlow
 import kotlinx.coroutines.launch
 import social.androiddev.domain.authentication.model.AuthStatus
 import social.androiddev.domain.authentication.usecase.GetAuthStatus
 import kotlin.coroutines.CoroutineContext
 
-interface RootComponentViewModel : InstanceKeeper.Instance {
-    val authState: StateFlow<AuthStatus?>
-}
-
-fun RootComponentViewModel(coroutineContext: CoroutineContext, getAuthStatus: GetAuthStatus): RootComponentViewModel =
-    RealRootComponentViewModel(coroutineContext, getAuthStatus)
-
-private class RealRootComponentViewModel(coroutineContext: CoroutineContext, getAuthStatus: GetAuthStatus) :
-    RootComponentViewModel {
+class RootComponentViewModel(coroutineContext: CoroutineContext, getAuthStatus: GetAuthStatus) :
+    InstanceKeeper.Instance {
     private val viewModelScope = CoroutineScope(coroutineContext + SupervisorJob())
 
-    private val _authState = MutableStateFlow<AuthStatus?>(null)
-    override val authState = _authState.asStateFlow()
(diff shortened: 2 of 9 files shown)


ui/root/src/commonMain/kotlin/social/androiddev/root/root/RootComponent.kt
```diff
@@ -49,4 +48,10 @@ interface RootComponent {
     sealed interface DeepLink {
         object None : DeepLink
     }
+
+    sealed class UiAuthStatus {
```
crocsandcoffee, line 52: Please move this to the same file as the composable or to its own file. It doesn't need to be coupled with the component file namespace

crocsandcoffee: Also is unknown ever used?

ui/root/src/commonMain/kotlin/social/androiddev/root/root/RootComponentViewModel.kt
```diff
@@ -17,31 +17,23 @@ import kotlinx.coroutines.CoroutineScope
...
-interface RootComponentViewModel : InstanceKeeper.Instance {
-    val authState: StateFlow<AuthStatus?>
-}
-
-fun RootComponentViewModel(coroutineContext: CoroutineContext, getAuthStatus: GetAuthStatus): RootComponentViewModel =
-    RealRootComponentViewModel(coroutineContext, getAuthStatus)
-
-private class RealRootComponentViewModel(coroutineContext: CoroutineContext, getAuthStatus: GetAuthStatus) :
-    RootComponentViewModel {
+class RootComponentViewModel(coroutineContext: CoroutineContext, getAuthStatus: GetAuthStatus) :
```
crocsandcoffee, line 26: Nit: please format so params are on their own lines for clearer readability/style/formatting

ui/root/src/commonMain/kotlin/social/androiddev/root/splash/SplashContent.kt
```diff
@@ -60,12 +60,12 @@ fun SplashContent(
         Text("Loading")
 
         LaunchedEffect(authStatus) {
-            if (authStatus is AuthStatus.Authorized) {
```
crocsandcoffee, line 63: We should always prefer exhaustive `when` over if/else chaining when using enum or sealed class

ui/root/src/commonMain/kotlin/social/androiddev/root/root/RootComponentViewModel.kt
```diff
@@ -17,31 +17,23 @@ import kotlinx.coroutines.CoroutineScope
...
+    InstanceKeeper.Instance {
     private val viewModelScope = CoroutineScope(coroutineContext + SupervisorJob())
 
-    private val _authState = MutableStateFlow<AuthStatus?>(null)
-    override val authState = _authState.asStateFlow()
+    private val _authState = MutableStateFlow<RootComponent.UiAuthStatus>(RootComponent.UiAuthStatus.Unknown)
```
crocsandcoffee, line 30: You can still simplify the `init` block and consolidate this private and public property using my suggestion from your other PR. Here's the link to my comment
https://github.com/AndroidDev-social/DodoForMastodon/pull/111#discussion_r1061676320

crocsandcoffee (commented):
Thanks for the follow-up!

ui/root/src/commonMain/kotlin/social/androiddev/root/root/RootComponentViewModel.kt
krizzu: the `stateIn` call is suspend call, so can't do it that way

ui/root/src/commonMain/kotlin/social/androiddev/root/root/RootComponent.kt
krizzu: Sure thing. Yes, `unknown` is used as initial value

ui/root/src/commonMain/kotlin/social/androiddev/root/root/RootComponentViewModel.kt
crocsandcoffee: I'm not sure I'm following... Please see the doc
https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/state-in.html

crocsandcoffee: Your repo returns a Flow.
You should be able to do something similar to 
```kotlin
val state: StateFlow<State> = repo.getFlow().stateIn(scope, SharingStarted.Eagerly, State.LOADING)
```

ui/root/src/commonMain/kotlin/social/androiddev/root/root/RootComponent.kt
crocsandcoffee: Ah right, I missed that. Might be clearer to rename to UIAuthStatus.Loading or make nullable instead.

ui/root/src/commonMain/kotlin/social/androiddev/root/root/RootComponentViewModel.kt
krizzu: That'd be true, if `stateIn` would not be a suspend call - The docs mention the function signature (at the bottom).

> Starts the upstream flow in a given scope, suspends until the first value is emitted, and returns a hotStateFlow of future emissions

The function signature is 
```kotlin
suspend fun <T> Flow<T>.stateIn(scope: CoroutineScope): StateFlow<T>
```

ui/root/src/commonMain/kotlin/social/androiddev/root/root/RootComponent.kt
krizzu: I like `Loading` better 👍

ui/root/src/commonMain/kotlin/social/androiddev/root/root/RootComponentViewModel.kt
krizzu: Will do - too bad spotless can't do that 😞

ui/root/src/commonMain/kotlin/social/androiddev/root/splash/SplashContent.kt
krizzu: I'm only interested in 2 cases out of 3, does the rule still apply?

crocsandcoffee: Still good practice, so you get the compiler warning/error if you ever add a new value to the enum/sealed class. You can always have it just be a no-op here for the ones you don't care about

ui/root/src/commonMain/kotlin/social/androiddev/root/root/RootComponentViewModel.kt
krizzu: Ah, perfect - That'd work!

krizzu:
@crocsandcoffee Thanks for checking it out!

ui/root/src/commonMain/kotlin/social/androiddev/root/splash/SplashContent.kt
```diff
@@ -60,13 +60,14 @@ fun SplashContent(
...
+            when (authStatus) {
                 // TODO(krzysztof): Do not navigate to timeline, until logout func is implemented
                 //                  https://github.com/AndroidDev-social/DodoForMastodon/issues/107
-                navigateToWelcome()
-                // navigateToTimeline()
-            } else if (authStatus is AuthStatus.Unauthorized) {
-                navigateToWelcome()
+                is UiAuthStatus.Authorized,
```
crocsandcoffee, line 66: Shouldn't you go to the timeline if you're authorized?

crocsandcoffee (approved):
Approving but plz confirm that last comment about navigating to timeline. Look good otherwise 👍

PR #125, opened 2023-01-11
Settings screens

### LINKED ISSUE #107: [FEAT]: Settings and Logout

krizzu:
## 📑 What does this PR do?

Added a Settings page, where user can logout from currently signed in server.

closes #107 

# ✅ Checklist

Run the App, log in and logout.

https://user-images.githubusercontent.com/6444719/211790910-9514d431-d0de-4d2c-987c-88642c0bbed8.mov

### DIFF
--- a/ui/root/src/commonMain/kotlin/social/androiddev/root/splash/DefaultSplashComponent.kt
+++ b/ui/root/src/commonMain/kotlin/social/androiddev/root/splash/DefaultSplashComponent.kt
@@ -16,15 +16,4 @@ import com.arkivanov.decompose.ComponentContext
 
 class DefaultSplashComponent(
     private val componentContext: ComponentContext,
-    private val navigateToLandingInternal: () -> Unit,
-    private val navigateToTimelineInternal: () -> Unit,
-) : SplashComponent, ComponentContext by componentContext {
-
-    override fun navigateToTimeline() {
-        navigateToTimelineInternal()
-    }
-
-    override fun navigateToLanding() {
-        navigateToLandingInternal()
-    }
-}
+) : SplashComponent, ComponentContext by componentContext
--- a/ui/settings/src/commonMain/kotlin/social/androiddev/settings/SettingsContent.kt
+++ b/ui/settings/src/commonMain/kotlin/social/androiddev/settings/SettingsContent.kt
@@ -13,13 +13,10 @@
 package social.androiddev.settings
 
 import androidx.compose.foundation.layout.Box
-import androidx.compose.foundation.layout.fillMaxSize
-import androidx.compose.foundation.layout.padding
 import androidx.compose.material.MaterialTheme
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
-import androidx.compose.ui.unit.dp
 import social.androiddev.common.composables.buttons.DodoButton
 import social.androiddev.common.composables.buttons.buttonColors
 
@@ -40,7 +37,7 @@ internal fun SettingsContent(
     modifier: Modifier = Modifier,
 ) {
     Box(
-        modifier = modifier.fillMaxSize().padding(bottom = 64.dp),
+        modifier = modifier,
         contentAlignment = Alignment.BottomCenter
     ) {
         DodoButton(
--- a/ui/settings/src/commonMain/kotlin/social/androiddev/settings/SettingsContent.kt
+++ b/ui/settings/src/commonMain/kotlin/social/androiddev/settings/SettingsContent.kt
@@ -39,7 +39,7 @@ internal fun SettingsContent(
 ) {
     Box(
         modifier = modifier.fillMaxSize(),
-        contentAlignment = Alignment.Center
+        contentAlignment = Alignment.BottomCenter
     ) {
         DodoButton(
             text = "Logout",
--- a/ui/settings/src/commonMain/kotlin/social/androiddev/settings/SettingsContent.kt
+++ b/ui/settings/src/commonMain/kotlin/social/androiddev/settings/SettingsContent.kt
@@ -1,0 +1,50 @@
+/*
+ * This file is part of Dodo.
+ *
+ * Dodo is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
+ *
+ * Dodo is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty
+ * of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along with Dodo.
+ * If not, see <https://www.gnu.org/licenses/>.
+ */
+package social.androiddev.settings
+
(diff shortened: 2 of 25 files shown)


ui/root/src/commonMain/kotlin/social/androiddev/root/splash/DefaultSplashComponent.kt
```diff
@@ -16,15 +16,4 @@ import com.arkivanov.decompose.ComponentContext
 
 class DefaultSplashComponent(
```
krizzu, line 17: I left a splash component just in case it'll be need in future, but I can drop it if required

crocsandcoffee: We can leave it for now in case we need to do kick off some task before taking the user to the first screen.

ui/settings/src/commonMain/kotlin/social/androiddev/settings/SettingsContent.kt
```diff
@@ -0,0 +1,52 @@
...
+    )
+}
+
+@Composable
+internal fun SettingsContent(
+    onLogout: () -> Unit,
+    modifier: Modifier = Modifier,
+) {
+    Box(
+        modifier = modifier.fillMaxSize().padding(bottom = 64.dp),
```
crocsandcoffee, line 43:
```suggestion
        modifier = modifier,
```

crocsandcoffee: `fillMaxSize` here would just override whatever size constraint was set by the caller. We should respect that size or use `then` modifier to only apply the size constraint if one wasn't specified. Same thing with padding

ui/settings/src/commonMain/kotlin/social/androiddev/settings/SettingsContent.kt
```diff
@@ -0,0 +1,52 @@
...
+        contentAlignment = Alignment.BottomCenter
+    ) {
+        DodoButton(
+            text = "Logout",
```
crocsandcoffee, line 47: This reminds me - we need to circle back to figuring out how we can use shared string resources across platforms instead of hard-coding

ui/signed-in/src/commonMain/kotlin/social/androiddev/signedin/composables/SignedInRootContent.kt
```diff
@@ -38,12 +49,43 @@ fun SignedInRootContent(
...
+                    {
+                        IconButton(onClick = component::navigateBack) {
+                            Icon(Icons.Default.ArrowBack, null)
+                        }
+                    }
+                    )
+            } else {
+                null
+            }
+        )
```
crocsandcoffee, line 88: here's a handy util funtion to add to the project
```kotlin
/**
 * Only return the result of [block] if [expression] is true, otherwise always returns null
 */
inline fun <T> nullUnless(expression: Boolean, block: () -> T): T? {
    contract {
        returnsNotNull() implies (expression)
        callsInPlace(block, InvocationKind.AT_MOST_ONCE)
    }
    return if (!expression) {
        null
    } else {
        block()
    }
}
```

and then this could be more readable and simplified to:
```kotlin
navigationIcon = nullUnless(childStack.backStack.isNotEmpty()) {
   {
         IconButton(onClick = component::navigateBack) {
               Icon(Icons.Default.ArrowBack, null)
         }
   }
}
```

crocsandcoffee (commented):
Looks good! Some minor comments and suggestions

ui/settings/src/commonMain/kotlin/social/androiddev/settings/SettingsContent.kt
crocsandcoffee: - [ ] Fix ignoring outside/caller sizing constraints

ui/signed-in/src/commonMain/kotlin/social/androiddev/signedin/composables/SignedInRootContent.kt
krizzu: Way better.
Any suggestion in which module the function should live? We have some util in `network`, but I imagine the function could be used in most places

ui/settings/src/commonMain/kotlin/social/androiddev/settings/SettingsContent.kt
krizzu: https://github.com/AndroidDev-social/DodoForMastodon/issues/126

ui/settings/src/commonMain/kotlin/social/androiddev/settings/SettingsContent.kt
krizzu: `fillMaxSize` does respect the incoming `maxWidth`/`maxHeight` constraint from the parent - it just fills up available space, instead of modifying it. Is that what you meant or am I missing something?

ui/signed-in/src/commonMain/kotlin/social/androiddev/signedin/composables/SignedInRootContent.kt
crocsandcoffee: Probably we should create a `kotlin-util` module, and any/all modules can depend on it if they want to. I have some more kotlin util functions I use that I will probably add. 

(Also runCatchingIgnoreCancelled` is one I already added to this project and should/could probably move to this new kotlin-util module as well

ui/settings/src/commonMain/kotlin/social/androiddev/settings/SettingsContent.kt
crocsandcoffee: Here's the scenario I am describing:

Given your `SettingsContent` internally takes in the passed in`modifier` and applies `.fillMaxSize().padding(bottom = 64.dp)`....

If the caller were to pass in `modifier.height(240.dp).width(120.dp)` or any other similar sizing modifier that _limits_ the size, your `.fillMaxSize` will effectively override that and force the composable to always take the full available space. 

You can quickly validate these via some preview functions

```kotlin

/**
* This should set the result size to fill maximum dimensions
*/

@Preview
@Composable
fun PreviewUseDefault() {
   SettingsContent(
      onLogout = {}.
   )
}

/**
* This should set the result size to h=240, w=120, but it will instead fill maximum dimensions
*/
@Preview
@Composable
fun PreviewThisShouldOverrideDefault() {
   SettingsContent(
      onLogout = {}.
      modifier = Modifier.height(240.dp).width(120.dp)
   )
}

```

krizzu: That makes sense! Indeed, `height`, `width` and `size` modifiers allow being overridden by incoming constraints (to be smaller or bigger), while `requiredHeight`, `requiredWidth`, and `requiredSize` set constraints to be exactly to the provided value.
Thanks for pointing this out

crocsandcoffee (approved):
🙌

